PR #22, opened 2025-05-13
Added usb enumeration in device mode scripts

kaushkir:
Added scripts to check USB enumeration along with readme file.

For windows - usb_enum_win.py
For linux - usb_enum_linux.py


mwasilew:
I don't know what to do with this script. Is it a part of the suite? @smuppand any idea how does this fit in the repository?

PR #32, opened 2025-05-14
[WIP] ci: add ShellCheck workflow for shell script linting

smuppand:
- Introduced a GitHub Actions workflow to run ShellCheck on all .sh files
- Ensures shell scripts follow best practices and common standards
- Helps catch syntax errors, quoting issues, and style problems early

This improves code quality and enforces consistency in shell script contributions.

```
@mwasilew @craigez @vnarapar 
```

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,14 @@
+name: Shell Lint
+
+on: [pull_request, push]
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install ShellCheck 0.9.0-1
+        run: sudo apt-get install -y shellcheck=0.9.0-1
+      - name: Run ShellCheck with exclusions
+        run: |
+          find . -name '*.sh' -print0 | xargs -0 shellcheck -S warning -e SC1091,SC2230,SC3043


.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install ShellCheck
+        run: sudo apt-get install -y shellcheck
+      - name: Run ShellCheck
+        run: |
+          find . -name '*.sh' -exec shellcheck {} +
```
mwasilew, line 14: we should probably exclude some codes. In test definitions the following codes are excluded: `SC1091 SC2230 SC3043`

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,15 @@
...
+
+on: [pull_request, push]
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install ShellCheck
+        run: sudo apt-get install -y shellcheck
```
mwasilew, line 11: maybe pin a version? Shellcheck behaviour changes in unpredicted way after upgrade. I'm currently using 0.9.0-1

mwasilew (changes requested):

mwasilew:
I think all checks should be green. If you want to introduce shellcheck you need to either fix all warnings beforehand or in the same PR.

.github/workflows/shellcheck.yml
smuppand: apt-get may not retrieve this version. Do you recommend downloading version 0.9.0-1 and installing it manually to ensure it is available for use.?

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install ShellCheck 0.9.0
+        run: sudo apt-get install -y shellcheck
+      - name: Run ShellCheck with exclusions
+        run: |
+          find . -name '*.sh' -print0 | xargs -0 shellcheck -e SC1091,SC2230,SC3043
```
mwasilew, line 14: please set level to `warning` so we reduce the number of complains to the most severe ones.
```
shellcheck -S warning
```

.github/workflows/shellcheck.yml
mwasilew:
```
apt-get install shellcheck=0.9.0-1
```
works on ubuntu 24.04 which is github's "latest"

.github/workflows/shellcheck.yml
smuppand: addressed with latest commit `aba7a6f74569cc6723a81820fc40912dd6d0ddf5`

.github/workflows/shellcheck.yml
smuppand: addressed with this commit `aba7a6f74569cc6723a81820fc40912dd6d0ddf5`

mwasilew (approved):

PR #34, opened 2025-05-15
Fix(scripts): restore execute permissions on shell scripts

smuppand:
This commit restores the missing execute (+x) permissions on shell scripts that were unintentionally modified in an earlier commit. These scripts are meant to be executable as part of the test harness and validation flows.
No other changes are introduced in this PR.

### Summary of changes:

- Reapplied chmod +x to affected *.sh files under this repo and related directories.

This fixes usability issues caused by prior 100755 → 100644 permission changes.

```
@mwasilew 
```


mwasilew:
could you remove the commits about autotools from this PR so we can merge only the mode change?

smuppand:
> could you remove the commits about autotools from this PR so we can merge only the mode change?

@mwasilew the changes are now made by dropping the autotools.

PR #42, opened 2025-05-20
Fix KernelCI premerge test plan

mwasilew:
Script calls and result reporting calls were not matching. This patch adds missing script calls and fixes paths for result files.


smuppand (approved):
LGTM

vnarapar (approved):
LGTM

mwasilew:
Rebased on top of current main

PR #43, opened 2025-05-21
Use current directory in case suite is not installed

mwasilew:
When suite is run from a different directory than /var/ functions won't be able to find tests by name. This patch fixes the issue and allows to discover tests that are executed from current path.


mwasilew:
Example test job with this fix https://lava.infra.foundries.io/scheduler/job/75440

PR #45, opened 2025-05-23
Add scripts to test USB Device enumeration on Host PCs

achauras-qcom:
Added test scripts to verify successful USB enumeration with user-input PID when DUT is connected to Host PC.

### DIFF
--- a/USBDevice/Linux/run.py
+++ b/USBDevice/Linux/run.py
@@ -1,0 +1,20 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+import os
+
+def check_enum(pid):
+	command = f'lsusb | grep -i {pid}'
+	if(os.system(command) == 0):
+		print("USB is properly enumerated")
+        return True
+	else:
+		print("USB enumeration failed!")
+        return False
+
+if __name__ == "__main__":
+    pid = input("Enter USB PID to check enumeration: ")
+    if check_enum(pid):
+        print("[PASS] USB Device Enumeration in {} is successful".format(pid))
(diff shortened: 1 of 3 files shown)


achauras-qcom:
@smuppand @mwasilew Please help review this PR.

smuppand:
@achauras-qcom As far as I know, this host-based test isn't applicable for Lava. It might become relevant once we migrate to Axiom. Would you prefer to wait until the migration is complete? @mwasilew  could you please confirm whether this test works on Lava?

mwasilew:
It might work in LAVA with container, but I don't know how useful this is. It tests the host rather than the device. In addition, there is no test script here, just a script that should be called with specific parameter. I'll try do demo this in LAVA if I have some time later today

achauras-qcom:
Thanks @mwasilew for the example. I'll try incorporating it.
I can keep shell script for linux if it's preferrable. I had tried out pyusb but was facing an issue with Windows. Once it's resolved, I will update the pyusb implementation so we have only 1 script.

achauras-qcom:
Regarding the use of the script for Host PC: Since USB is a master-slave protocol where master drives all communication, we need to test successful enumeration from master/host perspective only, which is the PC in this case. Our Axiom test scripts do the same.

mwasilew:
As far as I understand pyusb works on Linux, Windows and Mac. Do you have any logs/traces of the issue you're facing?

The builds we currently use don't include any gadged/adb drivers, so this test isn't very useful. It may change later though.

achauras-qcom:
In Windows, usb.core.find(find_all=True) is returning an empty list, probably due to missing libusb drivers.

If current builds don't include usb gadget drivers, then sure we can park this change for later. I will meanwhile work on resolving issues with my current script.

PR #46, opened 2025-05-23
Public CI Video Decode and Encode Shell Scripts.

abbajaj806:
Implemented shell scripts for public CI for Decode and Encode Scenarios.

Added Video encode test script
Added Video decode test script
Added readme document for video validation
Implemented Function for ActiveIP
Implemented Function for wget pull resource
Implemented Function for untar of resource
Implemented Function to check if file is already present or not

### DIFF
--- a/Runner/utils/functestlib.sh
+++ b/Runner/utils/functestlib.sh
@@ -21,7 +21,11 @@ log() {
 # Find test case path by name
 find_test_case_by_name() {
     local test_name="$1"
-    find $__RUNNER_SUITES_DIR -type d -iname "$test_name" 2>/dev/null
+    if [ -d "$__RUNNER_SUITES_DIR" ]; then
+        find $__RUNNER_SUITES_DIR -type d -iname "$test_name" 2>/dev/null
+    else
+        find "${PWD}" -type d -iname "$test_name" 2>/dev/null
+    fi
 }
 
 # Find test case path by name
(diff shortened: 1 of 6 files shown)


Runner/utils/functestlib.sh
```diff
@@ -91,3 +91,91 @@ functestlibdoc()
...
+extract_tar_from_url() {
+    local url="$1"
+    local filename
+    local extracted_files
+	
+	# Extract the filename from the URL
+    filename=$(basename "$url")
+	if check_tar_file "$filename"; then
+	   echo "[PASS] file already exists, Hence skipping downloading"
+	   return 1
```
mwasilew, line 128: The return codes are really confusing. Sometimes `1` means OK and sometimes it means failure. I propose to return `0` always when there is no problem and use non-zero codes when there are issues.

Runner/utils/functestlib.sh
```diff
@@ -91,3 +91,91 @@ functestlibdoc()
...
+ 
+    # Get first active IPv4 address (excluding loopback)
+    ip_addr=$(ip -4 addr show scope global up | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | head -n 1)
+ 
+    if [ -n "$ip_addr" ]; then
+        echo "[PASS] Network is active. IP address: $ip_addr"
+ 
+        if ping -c 1 -W 2 8.8.8.8 >/dev/null 2>&1; then
+            echo "[PASS] Internet is reachable."
+            return 1
```
mwasilew, line 107: Return codes in this function are reversed.

abbajaj806:
Some Rebase issue. Will raise another PR

PR #58, opened 2025-05-30
Fix to get the overall result in LAVA

vnarapar:
This commit is to update the overall pass/fail status of the LAVA job based on the test results

Added result_parse.sh

### DIFF
--- a/Runner/utils/result_parse.sh
+++ b/Runner/utils/result_parse.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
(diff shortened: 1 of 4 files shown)


Runner/utils/result_parse.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+find / -type f -name "*.res" 2>/dev/null | while IFS= read res_file; do
```
smuppand, line 6: Better accept a target directory (or defaults to current) walks recursively to find all .res files. Something like
`RESULTS_DIR="${1:-.}"`
`find "$RESULTS_DIR" -type f -name '*.res'`

PR #63, opened 2025-06-03
Add Probe_Failure_Check test to detect driver probe issues in dmesg

smuppand:
This patch introduces a new test case `Probe_Failure_Check` under the Kernel FunctionalArea suite. The script scans kernel logs (`journalctl` or `dmesg`) for signs of probe failures including:
 
- "probe failed"
- "failed to probe"
- "probe error"
 
Key Features:
- Compatible with both `journalctl` and `dmesg`-based systems
- Logs detailed output to `probe_failures.log`
- Structured to align with `[functestlib.sh](http://functestlib.sh/)` logging and result reporting
- Designed to run standalone or integrated in CI
 
This helps catch driver initialization issues that may otherwise go unnoticed in boot logs.

```
@vnarapar 
```

### DIFF
--- a/Runner/suites/Kernel/FunctionalArea/baseport/Probe_Failure_Check/run.sh
+++ b/Runner/suites/Kernel/FunctionalArea/baseport/Probe_Failure_Check/run.sh
@@ -1,0 +1,59 @@
+#!/bin/sh
+
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# Robustly source init_env and functestlib.sh
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+INIT_ENV=""
+SEARCH="$SCRIPT_DIR"
+while [ "$SEARCH" != "/" ]; do
+    if [ -f "$SEARCH/init_env" ]; then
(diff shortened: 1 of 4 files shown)


smuppand:
We are removing duplicate copyright from the root README.md file. Still the footer carries the necessary copyright.

Runner/suites/Kernel/FunctionalArea/baseport/Probe_Failure_Check/run.sh
```diff
@@ -0,0 +1,73 @@
...
+cd "$test_path" || exit 1
+
+res_file="./$TESTNAME.res"
+log_file="./probe_failures.log"
+
+log_info "-----------------------------------------------------------------------------------------"
+log_info "------------------- Starting $TESTNAME Testcase ----------------------------"
+
+# Helper: Fetch kernel logs
+get_kernel_log() {
```
vnarapar, line 42: Is this required to be added in common functions?

vnarapar:
Everything else looks good

Runner/suites/Kernel/FunctionalArea/baseport/Probe_Failure_Check/run.sh
smuppand: addressed with latest commit.

PR #66, opened 2025-06-04
Removed checks for Modem PIL subsystem

vnarapar:
This commit fixes the issue with remoteproc test by removing the checks for modem as it is not supported for QCS6490


smuppand (approved):
LGTM

PR #77, opened 2025-06-06
Add WiFi Connectivity Validation with Hybrid nmcli and wpa_supplicant Support

smuppand:
This patch adds a new hardware validation test for WiFi connectivity under the FunctionalArea suite. Key features include:
 
- Hybrid connection method:
  - Primary: nmcli (NetworkManager CLI)
  - Fallback: wpa_supplicant + udhcpc
- Supports reading SSID/password from:
  - CLI arguments
  - Environment variables
  - Fallback file (ssid_list.txt)
- Validates network IP assignment and performs internet ping test
- Handles systemd-networkd.service check and recovery
- Includes cleanup of background processes (wpa_supplicant) and configuration files
 
This test is useful for CI environments where both NetworkManager and wpa_supplicant setups might exist across devices.

```
@mwasilew @vnarapar @qcom-anilyada 
```

### DIFF
--- a/Runner/suites/Connectivity/WiFi/run.sh
+++ b/Runner/suites/Connectivity/WiFi/run.sh
@@ -1,0 +1,136 @@
+#!/bin/sh
+
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# Robustly find and source init_env
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+INIT_ENV=""
+SEARCH="$SCRIPT_DIR"
+while [ "$SEARCH" != "/" ]; do
+    if [ -f "$SEARCH/init_env" ]; then
+        INIT_ENV="$SEARCH/init_env"
+        break
+    fi
+    SEARCH=$(dirname "$SEARCH")
+done
+
+if [ -z "$INIT_ENV" ]; then
+    echo "[ERROR] Could not find init_env (starting at $SCRIPT_DIR)" >&2
+    exit 1
+fi
+
+# shellcheck disable=SC1090
+if [ -z "$__INIT_ENV_LOADED" ]; then
+    . "$INIT_ENV"
+fi
+
+# shellcheck disable=SC1090,SC1091
+. "$TOOLS/functestlib.sh"
+
+TESTNAME="WiFi"
+res_file="./$TESTNAME.res"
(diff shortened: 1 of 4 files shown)


Runner/suites/Connectivity/WiFi/run.sh
```diff
@@ -0,0 +1,143 @@
...
+            cleanup
+            exit 0
+        else
+            log_fail "Ping test failed after nmcli connection"
+        fi
+    fi
+fi
+
+# Fallback to wpa_supplicant + udhcpc if both are available
+if command -v wpa_supplicant >/dev/null 2>&1 && command -v udhcpc >/dev/null 2>&1; then
```
ricardosalveti, line 102: This might cause conflict with network-manager, maybe also good to check if nmcli is not available?

smuppand: > This might cause conflict with network-manager, maybe also good to check if nmcli is not available?
Sure, let me add a logic before launching the fallback method. Reason behing not checking for nmcli in `check_dependencies`
We intentionally excluded nmcli from check_dependencies() because the script is designed to be flexible across both full and minimal builds. It automatically detects and uses the best available method — nmcli or a fallback to wpa_supplicant + udhcpc. Including nmcli in mandatory checks would cause unnecessary test skips in minimal builds where nmcli isn't expected.

smuppand: Addressed with new changes.

Runner/suites/Connectivity/WiFi/WiFi_Firmware_Driver/run.sh
```diff
@@ -0,0 +1,68 @@
...
+test_path=$(find_test_case_by_name "$TESTNAME")
+cd "$test_path" || exit 1
+
+log_info "--------------------------------------------------------------------------"
+log_info "-------------------Starting $TESTNAME Testcase----------------------------"
+log_info "=== Test Initialization ==="
+
+# Paths for both targets
+KODIAK_FW="/lib/firmware/ath11k/WCN6750/hw1.0/qcm6490/wpss.mbn"
+LEMANS_FW="/lib/firmware/ath11k/WCN6855/hw2.1/amss.bin"
```
ricardosalveti, line 41: This does not look like a stable way to identify the target at runtime, we can easily share an image that has both installed.

Probably better to check /proc/device-tree/model or similar.

vnarapar (approved):
LGTM

PR #79, opened 2025-06-11
testlib: Mark test as SKIP if dependencies are missing

smuppand:
This patch updates the check_dependencies function in [functestlib.sh](http://functestlib.sh/) to gracefully skip tests when required tools are missing. Instead of failing the test (and reporting as FAIL in CI/LAVA), the test is now marked as SKIP when one or more dependencies are not present in the environment.
 
Key changes:
- check_dependencies now checks for all required commands and logs a warning for each missing dependency.
- If any dependency is missing, the test is marked as SKIP in the .res file, log_skip is called, and the script exits with code 0.
- No changes are required to individual [run.sh](http://run.sh/) scripts; all existing calls to check_dependencies automatically benefit from this update.
- This improves CI/LAVA reporting by distinguishing between genuine test failures and infrastructure/setup issues.
 
This change will reduce noise from false negatives in test dashboards and make it clear when failures are due to missing dependencies rather than actual test or system issues.


Runner/utils/functestlib.sh
```diff
@@ -132,13 +132,21 @@ unload_kernel_module() {
...
             missing=1
+            missing_cmds="$missing_cmds $cmd"
         fi
     done
-    [ "$missing" -ne 0 ] && exit 1
+    if [ "$missing" -ne 0 ]; then
+        # Use existing TESTNAME variable if present, fallback if not
+        testname="${TESTNAME:-UnknownTest}"
+        log_skip "$testname SKIP: missing dependencies:$missing_cmds"
+        echo "$testname SKIP" > "./$testname.res"
```
mwasilew, line 147: I propose to omit this line if $TESTNAME is not set.

Runner/utils/functestlib.sh
```diff
@@ -132,13 +132,21 @@ unload_kernel_module() {
...
+        exit 0
```
mwasilew, line 148: `exit 1` ?

mwasilew (changes requested):

Runner/utils/functestlib.sh
smuppand: addressed with latest changes.

Runner/utils/functestlib.sh
mwasilew: that's correct as far as I can tell. In the LAVA test plan here we use `|| true` to avoid this problem. I'm fine leaving it like it is.

PR #80, opened 2025-06-11
Ethernet Test: Dynamic Interface Detection and Robust CI Validation

smuppand:
This PR enhances the Ethernet validation script:
 
- Automatically detects the active Ethernet interface (instead of hardcoding eth0), supporting both ip and ifconfig tools.
- Adds bring-up logic with retries and detailed logging for interface readiness.
- Improves ping test logic for more reliable connectivity validation.
- Adds comprehensive log output to aid CI debugging and result analysis.
 
These changes make the Ethernet test more portable and robust across diverse hardware and build environments.


Runner/suites/Connectivity/Ethernet/run.sh
```diff
@@ -3,90 +3,133 @@
...
-    ip link set "$IFACE" up
-    sleep "$SLEEP_SEC"
-    if ip link show "$IFACE" | grep -q "state UP"; then
-        log_info "$IFACE is UP"
-        break
+
+if [ -z "$ETH_IFACES" ]; then
+    log_warn "No Ethernet interfaces detected."
+    echo "No Ethernet interfaces detected." >> "$summary_file"
+    echo "$TESTNAME SKIP" > "$res_file"
```
vnarapar, line 56: is it not a failure if  /sys/class/net/ is not having any ethernet interfaces

Runner/suites/Connectivity/Ethernet/run.sh
```diff
@@ -3,90 +3,133 @@
...
+
+any_passed=0
+any_tested=0
+
+for iface in $ETH_IFACES; do
+    log_info "---- Testing interface: $iface ----"
+
+    if ! is_interface_up "$iface"; then
+        log_warn "$iface is DOWN, skipping"
+        echo "$iface: SKIP (down/no cable)" >> "$summary_file"
```
vnarapar, line 68: i see an entry to check the cable aswell.
Do you think it is required to add that aswell
sh-5.2# cat /sys/class/net/eth0/carrier
0
0 - if cable is not present

smuppand: The `summary_file` will still be available, it is not being removed. We can capture everything needed from network maintainers to log into the summary based on the requirements.

PR #93, opened 2025-06-19
suites: Multimedia/Graphics/KMSCube: fix TESTNAME value

mwasilew:
Fix value of TESTNAME variable in KMSCube test. The value of the variable in the script didn't match the value in the meta-qcom-premerge test plan.


smuppand (approved):
LGTM

PR #94, opened 2025-06-20
Enhance send-to-lava.sh: POSIX-safe, atomic LAVA signal emission

smuppand:
This PR enhances the LAVA signal emission process used in `[send-to-lava.sh]` by:
- Making the script strictly POSIX-compliant
- Collecting all valid result lines into a temporary signal buffer
- Emitting the buffer in one atomic `cat` operation to avoid race conditions with `dmesg`
- Handling malformed or noisy result lines safely
- Ensuring cleanup of temporary artifacts
 
These changes were validated on RB3GEN2 and confirmed to resolve earlier issues where LAVA signals were malformed due to interleaved dmesg output.


vnarapar:
Please check if qualcomm liscense needs to be added as we are modifying the script now

vnarapar (approved):
LGTM

PR #105, opened 2025-07-02
Add zstd archive to tagged releases

mwasilew:
zstd archive helps with running tests on space constrained systems. It's already supported in LAVA and tuxlava (template renderer). This patch will make it possible to use qcom-linux-testkit in tuxlava.

### DIFF
--- a/.github/workflows/zstd.yml
+++ b/.github/workflows/zstd.yml
@@ -1,0 +1,37 @@
+name: Zstd Archive Release
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'  # Matching tags like v1.1.0
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install zstd
+        run: sudo apt-get update && sudo apt-get install -y zstd
+
+      - name: Extract version info
+        id: version
+        run: |
+          TAG_NAME="${GITHUB_REF#refs/tags/}"
+          echo "tag_name=$TAG_NAME" >> $GITHUB_OUTPUT
+
+      - name: Create .tar.zst archives using git archive
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/zstd.yml
```diff
@@ -0,0 +1,40 @@
+name: Zstd Archive Release
+
+on:
+  push:
+    tags:
```
smuppand, line 5: It should either track all tags or specific patterns?

.github/workflows/zstd.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          TAG_NAME="${GITHUB_REF#refs/tags/}"
+          SHORT_TAG=$(echo "$TAG_NAME" | cut -d '.' -f 1,2)
+          echo "tag_name=$TAG_NAME" >> $GITHUB_OUTPUT
+          echo "short_tag=$SHORT_TAG" >> $GITHUB_OUTPUT
+
+      - name: Create zstd archive without prefix
+        run: |
+          cd ..
+          tar --exclude='./qcom-linux-testkit/.git' -I zstd -cf ${{ steps.version.outputs.tag_name }}.tar.zst qcom-linux-testkit
```
smuppand, line 28: Avoid archiving the working directory, instead use a clean git archive. 

`-I zstd` its better to control compression level? but if we adopt git archive, this becomes simpler.

mwasilew:
@smuppand what is the tag format you'll be using?

smuppand:
> @smuppand what is the tag format you'll be using?

We can follow SemVer tags. 

Git tag: v1.1.0

Archive filename: qcom-testkit-v1.1.0.zst

.github/workflows/zstd.yml
mwasilew: Fixed.

.github/workflows/zstd.yml
mwasilew: Fixed

smuppand (approved):
LGTM

PR #116, opened 2025-07-14
ci: fix ShellCheck workflow to run correctly on PRs and main

smuppand:
- Updated  to properly differentiate between PRs and direct pushes to
- On pull_request: only runs ShellCheck on changed  files to avoid false positives from main
- On push to main and manual dispatch: runs ShellCheck on all  files in the repository
- Ensures accurate and relevant linting results on both CI and developer PRs
- Fixes issue where shellcheck errors from main were incorrectly reported on unrelated PRs

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,7 +1,7 @@
 name: Shell Lint
 
 on:
-  pull_request_target:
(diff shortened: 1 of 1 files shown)


vnarapar (approved):
LGTM

PR #118, opened 2025-07-14
CI: Improve ShellCheck workflow to skip deleted files in PRs

smuppand:
- Updated shellcheck GitHub Actions workflow to exclude deleted .sh files during pull request linting using --diff-filter=d.
- Ensures ShellCheck runs only on existing, modified shell scripts in PRs.
- Retains full linting on push and manual workflow_dispatch events.
- Removed use of 'realpath' to avoid issues with non-existent files.


vnarapar (approved):
LGTM

PR #122, opened 2025-07-21
ci(zstd): update workflow to use glob tag matching and support manual dispatch

smuppand:
This patch enhances the Zstd Archive Release workflow by:
 
Switching tag filter to a glob (v*.*.*) instead of a regex, ensuring semver-style tags (e.g. v1.2.3) actually trigger the workflow.
 
Adding workflow_dispatch, so you can manually run the job from the Actions UI without pushing a tag.
 
Emitting both tag_name and short_tag in the version-extraction step:
 
tag_name retains the leading v (e.g. v1.2.3)
 
short_tag strips it (e.g. 1.2.3)

### DIFF
--- a/.github/workflows/zstd.yml
+++ b/.github/workflows/zstd.yml
@@ -3,11 +3,13 @@ name: Zstd Archive Release
 on:
   push:
     tags:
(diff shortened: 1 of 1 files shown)


abbajaj806 (approved):
LGTM

PR #125, opened 2025-07-21
ci(zstd): use official GitHub release actions instead of softprops

smuppand:
Switch from softprops/action-gh-release to the built-in actions/create-release and actions/upload-release-asset so that our workflow complies with the org’s allowed actions policy. 
Both .tar.zst artifacts are now created and uploaded using only GitHub-verified actions.

### DIFF
--- a/.github/workflows/zstd.yml
+++ b/.github/workflows/zstd.yml
@@ -2,16 +2,13 @@ name: Zstd Archive Release
 
 on:
   push:
-    # Match semver tags like v1.2.3
     tags:
       - 'v*.*.*'
-  # Allow manual runs from the Actions UI
   workflow_dispatch:
 
 jobs:
   release:
     runs-on: ubuntu-latest
-
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/zstd.yml
```diff
@@ -24,30 +21,41 @@ jobs:
...
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ${{ github.workspace }}/${{ steps.version.outputs.tag_name }}.tar.zst
+          asset_name: ${{ steps.version.outputs.tag_name }}.tar.zst
+          asset_content_type: application/zstd
+
+      - name: Upload short-tag .tar.zst
+        uses: actions/upload-release-asset@v1
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
```
vnarapar, line 58: why to upload same .tar.zst file with 2 names. i suppose one will be with v1.tar.zst and other with 1.tar.zst right

PR #127, opened 2025-07-21
ci(zstd): switch to PAT for release creation and asset upload

smuppand:
This update modifies the Zstd Archive Release workflow to use a Personal Access Token (PAT) instead of the default GITHUB_TOKEN, addressing SSO and credential restrictions:
Replaces GITHUB_TOKEN with secrets.PAT for both actions/create-release@v1 and actions/upload-release-asset@v1
Maintains semver tag matching (v*.*.*) and manual dispatch support
Ensures the workflow can successfully authenticate under the organization’s SSO policy without modifying repository-level permissions

### DIFF
--- a/.github/workflows/zstd.yml
+++ b/.github/workflows/zstd.yml
@@ -40,13 +40,14 @@ jobs:
       - name: Create GitHub Release
         id: create_release
         uses: actions/create-release@v1
+        env:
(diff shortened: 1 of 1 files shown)


abbajaj806 (approved):
LGTM

PR #128, opened 2025-07-21
ci(zstd): add debug step for upload URL and token validation

smuppand:
This PR enhances the Zstd Archive Release workflow by inserting a diagnostic step before asset upload to help troubleshoot authentication and endpoint issues:
Debug upload context: Echoes the upload_url from the release creation step and prints the length of GITHUB_TOKEN to confirm PAT injection.
Ensures the .tar.zst asset upload step has the correct endpoint and a non-empty token before attempting the upload.
Aids in pinpointing “Bad credentials” errors by validating both the release URL and the authentication token at runtime.


vnarapar (approved):
LGTM

PR #129, opened 2025-07-21
ci(zstd): update git-archive prefix to qcom-linux-testkit/

smuppand:
This change adjusts the top-level folder inside the generated .tar.zst so that, when extracted, it appears under qcom-linux-testkit/ rather than test-definitions/.
Before: archives unpacked into test-definitions/...
After: archives unpack into qcom-linux-testkit/...


vnarapar (approved):
LGTM

PR #136, opened 2025-07-28
Fix: ensure wpss_remoteproc.res file is always generated

smuppand:
- Replaced all fail_and_exit and early exit paths with proper .res file creation (PASS, FAIL, or SKIP)

- Made RES_FILE path absolute to avoid issues caused by cd or relative directory changes

- Added a final fallback check to ensure .res is created before exiting

- Ensures LAVA and test harnesses can always rely on result presence for this test

### DIFF
--- a/Runner/suites/Kernel/Baseport/wpss_remoteproc/run.sh
+++ b/Runner/suites/Kernel/Baseport/wpss_remoteproc/run.sh
@@ -2,6 +2,7 @@
 
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
 # SPDX-License-Identifier: BSD-3-Clause-Clear
+#
 # Robustly find and source init_env
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 INIT_ENV=""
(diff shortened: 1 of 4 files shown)


Runner/suites/Kernel/Baseport/wpss_remoteproc/run.sh
```diff
@@ -174,11 +175,13 @@ else
...
         if [ "$dt_says_present" -eq 1 ]; then
-            fail_and_exit "DT lists $FW but no remoteproc and ath11k not loaded"
+            log_fail "DT lists $FW but no remoteproc and ath11k not loaded"
+	    echo "$TESTNAME FAIL" >"$RES_FILE"
+	    exit 1
         fi
-        log_skip "$TESTNAME SKIP – neither remoteproc nor ath11k path available"
-        echo "$TESTNAME SKIP" >"$RES_FILE"
-        exit 0
+            log_skip "$TESTNAME SKIP – neither remoteproc nor ath11k path available"
```
vnarapar, line 182: Can u check indentation here

vnarapar:
can you remove from adsp_remoteproc and cdsp_remoteproc aswell

smuppand:
> can you remove fail_and_exit from adsp_remoteproc and cdsp_remoteproc aswell

As per your request, I have removed the `fail_and_exit` from both the `adsp` and `cdsp_remoteproc` tests.

vnarapar (approved):
LGTM

PR #137, opened 2025-08-08
Add PipeWire support to AudioRecord and AudioPlayback tests

tmoida:
- Updated AudioRecord and AudioPlayback scripts to support both PulseAudio and PipeWire
- Added AUDIO_BACKEND variable for backend selection
- Updated README files with usage instructions for both backends
- Modified LAVA test plan to run tests with both PulseAudio and PipeWire

### DIFF
--- a/Runner/suites/Multimedia/Audio/AudioPlayback/run.sh
+++ b/Runner/suites/Multimedia/Audio/AudioPlayback/run.sh
@@ -1,94 +1,282 @@
 #!/bin/sh
-
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 
-# --------- Robustly source init_env and functestlib.sh ----------
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+
+# ---- Source init_env & tools ----
 INIT_ENV=""
 SEARCH="$SCRIPT_DIR"
 while [ "$SEARCH" != "/" ]; do
-    if [ -f "$SEARCH/init_env" ]; then
-        INIT_ENV="$SEARCH/init_env"
-        break
-    fi
-    SEARCH=$(dirname "$SEARCH")
+  if [ -f "$SEARCH/init_env" ]; then INIT_ENV="$SEARCH/init_env"; break; fi
+  SEARCH=$(dirname "$SEARCH")
 done
-
-if [ -z "$INIT_ENV" ]; then
-    echo "[ERROR] Could not find init_env (starting at $SCRIPT_DIR)" >&2
-    exit 1
-fi
-
-if [ -z "$__INIT_ENV_LOADED" ]; then
-    # shellcheck disable=SC1090
-    . "$INIT_ENV"
-fi
+[ -z "$INIT_ENV" ] && echo "[ERROR] init_env not found" >&2 && exit 1
+# shellcheck disable=SC1090
+[ -z "$__INIT_ENV_LOADED" ] && . "$INIT_ENV"
 # shellcheck disable=SC1090,SC1091
 . "$TOOLS/functestlib.sh"
-# ---------------------------------------------------------------
+. "$TOOLS/audio_common.sh"
 
 TESTNAME="AudioPlayback"
-TESTBINARY="paplay"
-TAR_URL="https://github.com/qualcomm-linux/qcom-linux-testkit/releases/download/Pulse-Audio-Files-v1.0/AudioClips.tar.gz"
-PLAYBACK_CLIP="AudioClips/yesterday_48KHz.wav"
-AUDIO_DEVICE="low-latency0"
-LOGDIR="results/audioplayback"
-RESULT_FILE="$TESTNAME.res"
-
-test_path=$(find_test_case_by_name "$TESTNAME")
-cd "$test_path" || exit 1
-
-# Prepare logdir
+RES_FILE="./${TESTNAME}.res"
+LOGDIR="results/${TESTNAME}"
 mkdir -p "$LOGDIR"
-chmod -R 777 "$LOGDIR"
 
-log_info "------------------------------------------------------------"
-log_info "------------------- Starting $TESTNAME Testcase ------------"
+# ---- Assets ----
+AUDIO_TAR_URL="${AUDIO_TAR_URL:-https://github.com/qualcomm-linux/qcom-linux-testkit/releases/download/Pulse-Audio-Files-v1.0/AudioClips.tar.gz}"
+export AUDIO_TAR_URL
 
-log_info "Checking if dependency binary is available"
-check_dependencies "$TESTBINARY" pgrep grep timeout
+# ------------- Defaults / CLI -------------
+AUDIO_BACKEND=""
+SINK_CHOICE="${SINK_CHOICE:-speakers}" # speakers|null
+FORMATS="${FORMATS:-wav}"
+DURATIONS="${DURATIONS:-short}" # short|medium|long
+LOOPS="${LOOPS:-1}"
+TIMEOUT="${TIMEOUT:-0}" # 0 = no timeout (recommended)
+STRICT="${STRICT:-0}"
+DMESG_SCAN="${DMESG_SCAN:-1}"
+VERBOSE=0
+EXTRACT_AUDIO_ASSETS="${EXTRACT_AUDIO_ASSETS:-true}"
+JUNIT_OUT=""
 
-# Download/extract audio if not present
-if [ ! -f "$PLAYBACK_CLIP" ]; then
-    log_info "Audio clip not found, downloading..."
-    extract_tar_from_url "$TAR_URL" || {
-        log_fail "Failed to fetch/extract playback audio tarball"
-        echo "$TESTNAME FAIL" > "$RESULT_FILE"
-        exit 1
-    }
+usage() {
+  cat <<EOF
+Usage: $0 [options]
+  --backend {pipewire|pulseaudio}
+  --sink {speakers|null}
+  --formats "wav"
+  --durations "short|short medium|short medium long"
+  --loops N
+  --timeout SECS # set 0 to disable watchdog
+  --strict
+  --no-dmesg
+  --no-extract-assets
+  --junit FILE.xml
+  --verbose
+  --help
+EOF
+}
+
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --backend) AUDIO_BACKEND="$2"; shift 2 ;;
+    --sink) SINK_CHOICE="$2"; shift 2 ;;
+    --formats) FORMATS="$2"; shift 2 ;;
+    --durations) DURATIONS="$2"; shift 2 ;;
+    --loops) LOOPS="$2"; shift 2 ;;
+    --timeout) TIMEOUT="$2"; shift 2 ;;
+    --strict) STRICT=1; shift ;;
+    --no-dmesg) DMESG_SCAN=0; shift ;;
+    --no-extract-assets) EXTRACT_AUDIO_ASSETS=false; shift ;;
+    --junit) JUNIT_OUT="$2"; shift 2 ;;
+    --verbose) export VERBOSE=1; shift ;;
+    --help|-h) usage; exit 0 ;;
+    *) log_warn "Unknown option: $1"; shift ;;
+  esac
+done
+
+# Ensure we run from the testcase dir
+test_path="$(find_test_case_by_name "$TESTNAME" 2>/dev/null || echo "$SCRIPT_DIR")"
+cd "$test_path" || { log_error "cd failed: $test_path"; echo "$TESTNAME FAIL" >"$RES_FILE"; exit 1; }
+
+log_info "---------------- Starting $TESTNAME ----------------"
+log_info "SoC: $(soc_id)"
+log_info "Args: backend=${AUDIO_BACKEND:-auto} sink=$SINK_CHOICE loops=$LOOPS timeout=$TIMEOUT formats='$FORMATS' durations='$DURATIONS' strict=$STRICT dmesg=$DMESG_SCAN extract=$EXTRACT_AUDIO_ASSETS"
+
+# Resolve backend
+[ -z "$AUDIO_BACKEND" ] && AUDIO_BACKEND="$(detect_audio_backend)"
+if [ -z "$AUDIO_BACKEND" ]; then
+  log_skip "$TESTNAME SKIP - no audio backend running"
+  echo "$TESTNAME SKIP" >"$RES_FILE"; exit 0
+fi
+log_info "Using backend: $AUDIO_BACKEND"
+
+if ! check_audio_daemon "$AUDIO_BACKEND"; then
+  log_skip "$TESTNAME SKIP - backend not available: $AUDIO_BACKEND"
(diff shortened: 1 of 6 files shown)


smuppand (changes requested):
@tmoida Could you please squash the commits into a single one?

Runner/suites/Multimedia/Audio/AudioPlayback/run.sh
```diff
@@ -29,26 +29,43 @@ fi
...
 
+# Determine backend and binary
+AUDIO_BACKEND="${AUDIO_BACKEND:-pulseaudio}"  # Default to pulseaudio if not set
+
+case "$AUDIO_BACKEND" in
+    pulseaudio)
+        TESTBINARY="paplay"
+        PLAY_CMD="paplay \"$PLAYBACK_CLIP\" -d \"$AUDIO_DEVICE\""
+        ;;
+    pipewire)
```
smuppand, line 46: Great to expose timeouts/ volume/ loops via env. Add a comment for a valid value range (eg paply volume 0-65536, pw-play 0.0 - etc)

smuppand:
@tmoida , please squash your commits.

Runner/suites/Multimedia/Audio/AudioPlayback/run.sh
```diff
@@ -68,16 +108,19 @@ fi
 log_info "Playback clip present: $PLAYBACK_CLIP"
 
 # --- Capture logs BEFORE playback (for debugging) ---
-dmesg > "$LOGDIR/dmesg_before.log"
+get_kernel_log > "$LOGDIR/dmesg_before.log"
 
 # --- Start the Playback, capture output ---
-timeout 15s paplay "$PLAYBACK_CLIP" -d "$AUDIO_DEVICE" > "$LOGDIR/playback_stdout.log" 2>&1
+for i in $(seq 1 "$PLAYBACK_LOOPS"); do
```
smuppand, line 114: Could you review the shellint-reported errors regarding unused variables? https://github.com/qualcomm-linux/qcom-linux-testkit/actions/runs/16964336815/job/48455697686?pr=137

smuppand (changes requested):

tmoida:
Hi @smuppand,

I've squashed the commits and fixed the Shell Check warning by using the loop variable. Please let me know if there is anything else that needs to be addressed.

Also. I have validated these scripts on Yocto builds, and they work fine. However, on Mainline (Nightly sanity) builds, they are being skipped due to a missing timeout binary. I wanted to check: will the timeout binary be included in future builds, or should I update the script to avoid using it?

smuppand:
> Hi @smuppand,
> 
> I've squashed the commits and fixed the Shell Check warning by using the loop variable. Please let me know if there is anything else that needs to be addressed.
> 
> Also. I have validated these scripts on Yocto builds, and they work fine. However, on Mainline (Nightly sanity) builds, they are being skipped due to a missing timeout binary. I wanted to check: will the timeout binary be included in future builds, or should I update the script to avoid using it?

You can use the library function run_with_timeout() if the build does not include the timeout utility.

tmoida:
Hi @smuppand,

Thank you for your feedback.

I have implemented the requested modular structure and made the following changes:

```
1. Moved all core logic for playback and recording into a shared `libaudio.sh`.
2. Created minimal `run.sh` scripts under AudioPlayback and AudioRecord that only parse CLI flags and invoke functions from `libaudio.sh`.
3. Wrapped all binary calls (paplay, pw-play, parec, pw-record) inside `libaudio.sh`, eliminating direct calls in `run.sh`.
4. Added support for both CLI flags and environment variables for configuration.
```

**Why separate run.sh scripts for playback and record?**
Regarding the use of separate `run.sh` scripts for playback and recording: We are utilizing `run-test.sh` as the main test runner, which expects each test case to be in its own directory and produce a result file named after that directory (e.g., `AudioPlayback.res`, `AudioRecord.res`). Using a single `run.sh` within a common Audio/ directory would prevent `run-test.sh` from locating the correct `.res` file, potentially leading to false negatives, even when tests pass.

To address this and ensure compatibility with `run-test.sh`, we have maintained separate directories and minimal `run.sh` wrappers for each test case. This approach allows for correct result parsing while centralizing the logic in `libaudio.sh`.

All usage details, CLI options, and examples are documented in the respective `Read_me.md` files for easy reference.

Let me know if you’d like any further refinements!

Runner/suites/Multimedia/Audio/AudioPlayback/run.sh
```diff
@@ -1,94 +1,66 @@
...
-    fi
-    SEARCH=$(dirname "$SEARCH")
+  if [ -f "$SEARCH/init_env" ]; then
+    INIT_ENV="$SEARCH/init_env"
+    break
+  fi
+  SEARCH=$(dirname "$SEARCH")
 done
+[ -z "$INIT_ENV" ] && echo "[ERROR] init_env not found" && exit 1
+# shellcheck source=/dev/null
```
smuppand, line 18: using `# shellcheck source=/dev/null` is a blunt silencer that hides real problems and degrades analysis. It tells ShellCheck “pretend this sourced file exists but is empty, It masks real bugs.

Runner/suites/Multimedia/Audio/AudioPlayback/run.sh
```diff
@@ -1,94 +1,66 @@
...
-TESTBINARY="paplay"
-TAR_URL="https://github.com/qualcomm-linux/qcom-linux-testkit/releases/download/Pulse-Audio-Files-v1.0/AudioClips.tar.gz"
-PLAYBACK_CLIP="AudioClips/yesterday_48KHz.wav"
-AUDIO_DEVICE="low-latency0"
-LOGDIR="results/audioplayback"
-RESULT_FILE="$TESTNAME.res"
+. "$TOOLS/libaudio.sh"
 
-test_path=$(find_test_case_by_name "$TESTNAME")
-cd "$test_path" || exit 1
```
smuppand, line 40: This approach will fail if we run any test from the run-test root, and the .res file will be generated in the directory where the command is executed, rather than in the test directory. Please keep the expected snippet.

Runner/suites/Multimedia/Audio/AudioPlayback/run.sh
```diff
@@ -1,94 +1,66 @@
...
+# Override with BusyBox-compatible version
+check_network_status() {
```
smuppand, line 25: check_network_status override isn’t used—drop it to keep things tight. 
Use available built-in fiunctions like [get_ip_address()](https://github.com/qualcomm-linux/qcom-linux-testkit/blob/main/Runner/utils/functestlib.sh#L498) to avoid duplication.

Runner/suites/Multimedia/Audio/AudioRecord/run.sh
```diff
@@ -1,77 +1,62 @@
...
-    fi
-    SEARCH=$(dirname "$SEARCH")
+  if [ -f "$SEARCH/init_env" ]; then
+    INIT_ENV="$SEARCH/init_env"
+    break
+  fi
+  SEARCH=$(dirname "$SEARCH")
 done
+[ -z "$INIT_ENV" ] && echo "[ERROR] init_env not found" && exit 1
+# shellcheck source=/dev/null
```
smuppand, line 18: NIT

Runner/suites/Multimedia/Audio/AudioRecord/run.sh
```diff
@@ -1,77 +1,62 @@
...
-if [ -z "$INIT_ENV" ]; then
-    echo "[ERROR] Could not find init_env (starting at $SCRIPT_DIR)" >&2
-    exit 1
-fi
-
-if [ -z "$__INIT_ENV_LOADED" ]; then
-    # shellcheck disable=SC1090
-    . "$INIT_ENV"
-fi
-# shellcheck disable=SC1090,SC1091
```
smuppand, line 27: NIT

Runner/suites/Multimedia/Audio/AudioRecord/run.sh
```diff
@@ -1,77 +1,62 @@
...
+# Set TESTNAME
 TESTNAME="AudioRecord"
-TESTBINARY="parec"
-RECORD_FILE="/tmp/rec1.wav"
-AUDIO_DEVICE="regular0"
-LOGDIR="results/audiorecord"
-RESULT_FILE="$TESTNAME.res"
 
-test_path=$(find_test_case_by_name "$TESTNAME")
-cd "$test_path" || exit 1
```
smuppand, line 39: NIT

Runner/utils/libaudio.sh
```diff
@@ -0,0 +1,200 @@
...
+validate_file_exists() {
+    file="$1"
+    if [ ! -s "$file" ]; then
+        log_fail "File missing or empty: $file"
+        return 1
+    fi
+    return 0
+}
+
+execute_with_timeout() {
```
smuppand, line 52: why to introduce a new function instead of using run_with_timeout?

Runner/utils/libaudio.sh
```diff
@@ -0,0 +1,200 @@
...
+setup_playback_command() {
+    if [ "$AUDIO_BACKEND" = "pulseaudio" ]; then
+        echo "paplay --volume=$PLAYBACK_VOLUME \"$PLAYBACK_CLIP\" -d \"$AUDIO_PLAYBACK_DEVICE\""
+    elif [ "$AUDIO_BACKEND" = "pipewire" ]; then
+        echo "pw-play --volume=$PLAYBACK_VOLUME \"$PLAYBACK_CLIP\""
+    fi
+}
+
+audio_download_clip() {
+    TAR_URL="https://github.com/qualcomm-linux/qcom-linux-testkit/releases/download/Pulse-Audio-Files-v1.0/AudioClips.tar.gz"
```
smuppand, line 98: this should go to a file constant

Runner/utils/libaudio.sh
```diff
@@ -0,0 +1,200 @@
...
+		else
+			log_fail "Playback loop $i: Failed with error code $ret"
+			# shellcheck disable=SC2034
+			PLAYBACK_SUCCESS=1
+			break  # Early exit on first true failure
+		fi
+    done
+
+    get_kernel_log > "$LOGDIR/dmesg_after.log"
+    scan_dmesg_errors "audio" "$LOGDIR"
```
smuppand, line 143: Your function signature is scan_dmesg_errors(prefix, module_regex, [exclude]).
 
Current calls use scan_dmesg_errors "audio" "$LOGDIR" (reversed). This silently weakens the scan. It returns 0 when errors are found, 1 when clean. Runners should only gate FAIL on this if a strict flag is set (to avoid flakiness).

Runner/utils/libaudio.sh
```diff
@@ -0,0 +1,200 @@
...
+    check_dependencies "$([ "$AUDIO_BACKEND" = "pulseaudio" ] && echo "paplay" || echo "pw-play")" pgrep grep
+    audio_download_clip
+
+    get_kernel_log > "$LOGDIR/dmesg_before.log"
+    PLAY_CMD=$(setup_playback_command)
+    PLAY_SUCCESS=0
+
+    for i in $(seq 1 "$PLAYBACK_LOOPS"); do
+        log_info "Playback loop $i of $PLAYBACK_LOOPS"
+        execute_with_timeout "$PLAYBACK_TIMEOUT" sh -c "$PLAY_CMD" >> "$LOGDIR/playback_stdout.log" 2>&1
```
smuppand, line 126: Defaults like 15s are fine for timeout(1) but may break run_with_timeout (usually expects integers). Normalize once, inside execute_with_timeout.

Runner/utils/libaudio.sh
```diff
@@ -0,0 +1,200 @@
...
+}
+
+audio_playback() {
+    check_audio_daemon
+    check_dependencies "$([ "$AUDIO_BACKEND" = "pulseaudio" ] && echo "paplay" || echo "pw-play")" pgrep grep
+    audio_download_clip
+
+    get_kernel_log > "$LOGDIR/dmesg_before.log"
+    PLAY_CMD=$(setup_playback_command)
+    PLAY_SUCCESS=0
```
smuppand, line 122: Playback uses PLAY_SUCCESS but sets PLAYBACK_SUCCESS in one branch—typo causes false PASS.

Runner/utils/libaudio.sh
```diff
@@ -0,0 +1,200 @@
...
+    fi
+
+}
+
+audio_record() {
+    check_audio_daemon
+    if [ "$AUDIO_BACKEND" = "pipewire" ]; then
+        select_pipewire_source
+    fi
+    check_dependencies "$([ "$AUDIO_BACKEND" = "pulseaudio" ] && echo "parec" || echo "pw-record")" pgrep grep
```
smuppand, line 160: select_pipewire_source uses wpctl but dependency check doesn’t require it.
 
Minor: pgrep grep in check_dependencies is unnecessary; keep pgrep, drop grep.

Runner/utils/libaudio.sh
```diff
@@ -0,0 +1,200 @@
...
+    scan_dmesg_errors "audio" "$LOGDIR"
+
+    if [ "$PLAY_SUCCESS" -eq 0 ]; then
+        log_pass_exit "$TESTNAME" "Audio playback test completed successfully"
+        echo "$TESTNAME PASS" > "$RESULT_FILE"
+    else
+        log_fail_exit "$TESTNAME" "Audio playback test failed"
+        echo "$TESTNAME FAIL" > "$RESULT_FILE"
+    fi
+
```
smuppand, line 152: Transferring PASS/FAIL/SKIP ownership to the top-level runner. These shared libraries will be used across the audio files.

smuppand (changes requested):
Also woth adding 

- WAV 16-bit/48k, 44.1k; Ogg Vorbis/Opus (if tools available), AAC.
- Short (5s), medium (30s), long (120s) smoke tiers.
- Mixer dumps (pactl list, pw-dump) on failure.

Runner/utils/functestlib.sh
```diff
@@ -263,7 +263,7 @@ check_network_status() {
     echo "[INFO] Checking network connectivity..."
 
     # Get first active IPv4 address (excluding loopback)
-    ip_addr=$(ip -4 addr show scope global up | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | head -n 1)
+	ip_addr=$(ip -4 addr show scope global up | awk '/inet / {print $2}' | cut -d/ -f1 | head -n 1)
```
smuppand, line 266: AFAIK, this approach won't work if the routing table includes multiple NICs, VRFs, or asymmetric configurations. It might be better to add an interface-scan method to simply retrieve the first global address.

Runner/utils/audio_common.sh
```diff
@@ -0,0 +1,548 @@
+#!/bin/sh
+# SPDX-License-Identifier: BSD-3-Clause-Clear
```
smuppand, line 2: The copyright information is missing.

smuppand (changes requested):

Runner/utils/audio_common.sh
```diff
@@ -0,0 +1,551 @@
+#!/bin/sh
+
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
```
smuppand, line 5: Remove white space.

smuppand (approved):
LGTM

PR #144, opened 2025-08-21
adsp_remoteproc: enforce fixed 30s pre-stop

smuppand:
* Always wait a fixed PRE_STOP_DELAY (default 30s) before stopping ADSP.
* Take a quick audio runtime-PM snapshot before the delay; optionally gate the stop after the delay with FATAL_ON_UNSUSPENDED=1:
  - If enabled and the stack is not suspended/unsupported after the fixed delay, abort the test before issuing the stop.
* Print effective tunables at startup for clarity.
* Retain dynamic audio driver discovery (no hardcoding), single module status log, and whole-stack PM checks (platform sound, ALSA cards, SoundWire). Accept 'suspended' or 'unsupported' as OK.
* Preserve remoteproc stop/start sequencing, transition logs, and optional RPMsg ping.

Tunables (env)
--------------
* PRE_STOP_DELAY   : seconds to wait before stop (default: 30)
* FATAL_ON_UNSUSPENDED : 1 to abort if PM not OK after delay (default: 0)
* STOP_TO / START_TO / POLL_I remain unchanged.

Examples
--------
* Always wait 30s, abort if audio not suspended/unsupported afterwards FATAL_ON_UNSUSPENDED=1 ./run.sh
*  Use a longer fixed delay
PRE_STOP_DELAY=45 ./run.sh

### DIFF
--- a/Runner/suites/Kernel/Baseport/adsp_remoteproc/run.sh
+++ b/Runner/suites/Kernel/Baseport/adsp_remoteproc/run.sh
@@ -1,5 +1,4 @@
 #!/bin/sh
-
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 # Robustly find and source init_env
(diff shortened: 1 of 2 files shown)


Runner/suites/Kernel/Baseport/adsp_remoteproc/run.sh
```diff
@@ -39,42 +38,149 @@ log_info "----------------------------------------------------------------------
...
+        else
+            platform_mod=""
+        fi
+        break
+    done
+
+    CHECK_NODES=""
+    if [ -z "$platform_drv" ]; then
+        log_warn "No suitable platform audio driver found (module+bound); skipping suspend snapshot"
+        AUDIO_PM_SNAPSHOT_OK=1
```
vnarapar, line 78: can we append skip to result here

Runner/suites/Kernel/Baseport/adsp_remoteproc/run.sh
```diff
@@ -39,42 +38,149 @@ log_info "----------------------------------------------------------------------
...
+
+    # 3) SoundWire slaves
+    for f in /sys/bus/soundwire/devices/*/power/runtime_status; do
+        [ -f "$f" ] && CHECK_NODES="$CHECK_NODES $f"
+    done
 
-log_info "DEBUG: STOP_TO=$STOP_TO START_TO=$START_TO POLL_I=$POLL_I"
+    if [ -z "$CHECK_NODES" ]; then
+        log_warn "No runtime_status nodes found for audio stack; treating snapshot as OK"
+        AUDIO_PM_SNAPSHOT_OK=1
```
vnarapar, line 121: can u add skip so that it will show in the result

Runner/utils/functestlib.sh
```diff
@@ -1200,7 +1200,7 @@ dump_rproc_logs() {
     logfile="rproc_${base}_${label}_${ts}.log"
     log_info "Dumping ${base} [${label}] → ${logfile}"
     [ -r "$rpath/trace" ] && cat "$rpath/trace" >"$logfile"
-    dmesg --ctime | grep -i "$base" >>"$logfile" 2>/dev/null || :
+    dmesg | grep -i "$base" >>"$logfile" 2>/dev/null || :
```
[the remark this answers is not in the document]
smuppand: > why --ctime is removed here

The current dmesg utility does not include this option, so it will display a usage error.

Runner/suites/Kernel/Baseport/adsp_remoteproc/run.sh
smuppand: > can we append skip to result here

AFAIK, we cannot skip the test because the driver does not have the capability to display the PM stats, as it shows unsupported. The message reads: `Audio PM snapshot: OK (suspended/unsupported)`. Based on this, we are able to skip the entire test.

Runner/suites/Kernel/Baseport/adsp_remoteproc/run.sh
smuppand: > can u add skip so that it will show in the result

It's the same here. We can't add SKIP, so instead we're using log_info for debugging purposes.

PR #162, opened 2025-09-07
Added support for multiple arguments

vnarapar:
This change provides greater flexibility by allowing users to provide specific parameters to individual testcases from run-test.sh


Runner/run-test.sh
```diff
@@ -120,14 +123,22 @@ print_summary() {
 }
 
 if [ "$#" -eq 0 ]; then
-    log "Usage: $0 [all | <testcase_name>]"
+    echo "Usage:"
```
smuppand, line 126: Nit/UX: Please refactor the usage/help output
Right now we mix lines with several echos and print “Usage” twice. For CLI UX (and tooling), help should be a single, clean block printed to stderr, with a -h/--help path and clear exit codes.

smuppand (changes requested):

Runner/run-test.sh
vnarapar: Updated as per comment to address clean block with help when -h/--help is given and added exit codes as suggested

Runner/run-test.sh
```diff
@@ -119,15 +122,35 @@ print_summary() {
     log_info "=================================="
 }
 
-if [ "$#" -eq 0 ]; then
-    log "Usage: $0 [all | <testcase_name>]"
-    exit 1
+print_usage() {
+    cat >&2 <<EOF
+Usage:
+  $0 all
```
smuppand, line 128: `$0` prints the script exactly as it was invoked, which might include a long/ugly path (e.g. ./<test>/[run.sh](http://run.sh/) or /tmp/x/y/[run.sh](http://run.sh/)). That clutters usage text and varies across environments.
 
`"${0##*/}"` strips any leading path and prints just the basename (e.g. [run.sh](http://run.sh/)).

Runner/run-test.sh
```diff
@@ -119,15 +122,35 @@ print_summary() {
...
+  $0 <testcase_name> [arg1 arg2 ...]
```
smuppand, line 129: NIT

smuppand (changes requested):

PR #164, opened 2025-09-07
Change directory for qcrypto and kselftest

vnarapar:
This commit has minor fix to change the directory of qcrypto/kselftest


Runner/suites/Kernel/Baseport/qcom_crypto/README_qcom_crypto.md
```diff
@@ -1,83 +1,83 @@
...
+SKIPPED:
+ None
+[INFO] 1970-01-01 05:44:18 - ==================================
+```
+4. Results will be available in the `/<user-defined-location>/Runner/suites/Kernel/FunctionalArea/baseport/qcom_crypto/` directory.
+
+## Notes
+
+- It uses kcapi to validate AES-CBC encryption/decryption and HMAC-SHA256 hashing.
 - If any test fails, the script logs the expected vs actual output and exits with a failure code.
```
smuppand, line 83: Is there a particular reason for removing this link, since it is supposed to display a reason in case of failures?

Runner/suites/Kernel/Baseport/qcom_crypto/README_qcom_crypto.md
```diff
@@ -1,83 +1,83 @@
...
+
+- `kcapi` (Kernel Crypto tool) Available in /usr/bin
+- `libkcapi.so.1` (Kernel Crypto library) Available in /usr/lib
+
+## Directory Structure
+```
+Runner/
+├── suites/
+│   ├── Kernel/
+│   │   ├── FunctionalArea/
```
smuppand, line 29: The new folder structure should also be reflected here.

smuppand (changes requested):

Runner/suites/Kernel/Baseport/qcom_crypto/README_qcom_crypto.md
vnarapar: There is no difference, not sure why it is coming as diff

PR #167, opened 2025-09-16
Add modular hw-probe runner (Debian/Ubuntu + Docker) with install/update, offline save, upload, auto-extract, and network checks

smuppand:
This PR introduces a fully-modular hw-probe runner under Runner/tools/hw-probe/ with shared utilities in Runner/utils/. It supports Debian/Ubuntu native and Docker execution, handles install/update (incl. specific versions), works offline (save only) or online (upload), and can auto-extract the generated report. All logging and basic checks reuse our existing functestlib.sh (no duplicated infra).

Why

- Standardize hardware probing across environments.
- One-line UX to install deps, run probes, save artifacts, and (optionally) upload.
- Make behavior reliable in CI (network checks, explicit output paths, extraction).

What’s included

New CLI: Runner/tools/hw-probe/run.sh (POSIX sh)

- --mode local|docker

- --upload yes|no (auto-disables when offline)

- --out DIR (default ./hw-probe_out)

- --extract yes|no (auto-unpacks hw.info.txz to OUT/extracted-<ts>)

- --install, --version <ver>, --update, --list-versions, --deps-only

- --probe-args "<extra hw-probe flags>"

- --dry-run, --verbose

Shared libs (POSIX sh):

- utils/lib_common.sh – tiny helpers (root/sudo, ensure_dir, OS detect, network check via check_network_status)

- utils/lib_apt.sh – APT install/update/version helpers (guarded by network checks)

- utils/lib_docker.sh – docker presence/install/run; skip pulls when offline, require local image

- utils/lib_hwprobe.sh – install/update/run (local & docker), save with -save (correct flag), parse “Probe URL”, print saved path, auto-extract when requested

Docs: Runner/tools/hw-probe/README.md (usage, examples, artifacts, offline tips, Docker notes, troubleshooting)

Key details

Uses functestlib.sh for all logging (log_info, log_warn, log_error, log_fail, log_skip) and check_dependencies; prefers check_network_status when available.

No reimplementation of logging/deps logic; libs are minimal and shellcheck-friendly.

Corrects flag usage: we use hw-probe -save DIR (not -dump) so artifacts persist deterministically.

Local runs use sudo -E automatically when not root.

Docker runs mirror upstream’s guidance: --privileged --net=host --pid=host and read-only host mounts (/dev, /lib/modules, /etc/os-release, /var/log) plus -v OUT:/out to persist results.

If --upload yes but offline, runner downgrades to no for that run (logs a warning).

After each run, we print where the archive was saved and (if uploaded) the probe URL. With --extract yes, we unpack the .txz to OUT/extracted-<ts>.

Usage examples

# Local, save only, then auto-extract
./run.sh --mode local --upload no --extract yes

# Local, upload + extract
./run.sh --mode local --upload yes --extract yes

# Docker, save only, custom out dir
./run.sh --mode docker --upload no --out ./out --extract yes

# Ensure latest hw-probe (if missing) then run
./run.sh --install --mode local

# Install a specific version
./run.sh --install --version 1.6

# Update to latest
./run.sh --update

# List available versions
./run.sh --list-versions

Test plan

-  Debian/Ubuntu: local run with/without --upload

-  Offline mode: install/update guarded; --upload auto-disabled

-  Docker mode: image pull online; offline requires local image; artifacts saved under OUT

-  Auto-extraction: creates OUT/extracted-<ts> and unpacks hw.info.txz

-  ShellCheck: POSIX sh; dynamic-source warnings only (expected for sourcing)

-  Reuse of existing infra: logging/dep checks via functestlib.sh

### DIFF
--- a/Runner/utils/lib_common.sh
+++ b/Runner/utils/lib_common.sh
@@ -1,0 +1,131 @@
+#!/bin/sh
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+# POSIX helpers that defer to functestlib.sh where available.
+
+# --- fallbacks only if functestlib wasn't sourced ---
+if ! command -v log_info >/dev/null 2>&1; then
+  log_info() { printf '[INFO] %s\n' "$*"; }
+fi
+if ! command -v log_warn >/dev/null 2>&1; then
+  log_warn() { printf '[WARN] %s\n' "$*" >&2; }
+fi
+if ! command -v log_error >/dev/null 2>&1; then
+  log_error() { printf '[ERROR] %s\n' "$*" >&2; }
+fi
+if ! command -v log_fail >/dev/null 2>&1; then
+  log_fail() { printf '[FAIL] %s\n' "$*" >&2; }
+fi
+if ! command -v log_skip >/dev/null 2>&1; then
+  log_skip() { printf '[SKIP] %s\n' "$*"; }
+fi
+
+die() { log_error "$*"; exit 1; }
+
+# --- cmd exists ---
+have_cmd() { command -v "$1" >/dev/null 2>&1; }
+
+# --- root / sudo (non-interactive aware) ---
+SUDO=""
+is_root() { [ "$(id -u 2>/dev/null || echo 1)" -eq 0 ]; }
+
+# Returns:
+# 0 => we can escalate non-interactively (sets $SUDO or empty when already root)
+# 2 => escalation would PROMPT (caller should SKIP gracefully)
+ensure_root_noprompt() {
+  if is_root; then
+    SUDO=""
+    return 0
+  fi
+
+  if have_cmd sudo; then
+    if sudo -n true >/dev/null 2>&1; then
+      SUDO="sudo -E"
+      return 0
+    fi
+    if [ -n "${SUDO_ASKPASS:-}" ] && [ -x "${SUDO_ASKPASS:-/nonexistent}" ]; then
+      if sudo -A -n true >/dev/null 2>&1; then
+        SUDO="sudo -A -E"
+        return 0
+      fi
+    fi
+  fi
+
+  if have_cmd doas; then
+    if doas -n true >/dev/null 2>&1; then
+      SUDO="doas"
+      return 0
+    fi
+  fi
+
+  return 2
+}
+
+need_root() {
+  ensure_root_noprompt
+  rc=$?
+  if [ "$rc" -eq 0 ]; then
+    return 0
+  fi
+  die "Root privileges required but non-interactive sudo/doas is unavailable (would prompt). Re-run as root or configure passwordless sudo."
+}
+
+need_root_or_skip() {
+  ensure_root_noprompt
+  rc=$?
+  if [ "$rc" -eq 0 ]; then
+    return 0
+  fi
+  log_skip "Root required but sudo/doas would prompt (no passwordless method)."
+  return 2
+}
+
+# --- ensure dir ---
+ensure_dir() {
+  d="$1"
+  if [ -d "$d" ]; then
+    return 0
+  fi
+  mkdir -p "$d" 2>/dev/null && return 0
+  if need_root_or_skip; then
+    $SUDO mkdir -p "$d" || return 1
+    return 0
+  fi
+  return 1
+}
+
+# --- OS detect (Debian/Ubuntu) ---
+detect_os_like() {
+  if [ -r /etc/os-release ]; then
+    # shellcheck source=/dev/null
+    . /etc/os-release
+    printf '%s\n' "${ID_LIKE:-$ID}"
+    return 0
+  fi
+  printf '%s\n' "unknown"
+}
+
+is_debianish() {
+  like="$(detect_os_like | tr '[:upper:]' '[:lower:]')"
+  echo "$like" | grep -Eq 'debian|ubuntu'
+}
+
+# --- network check: prefer functestlib’s richer checks if present ---
+network_is_ok() {
+  if command -v check_network_status >/dev/null 2>&1; then
+    check_network_status
+    return $?
+  fi
+  ping -c 1 -W 1 1.1.1.1 >/dev/null 2>&1 || ping -c 1 -W 1 8.8.8.8 >/dev/null 2>&1
+}
+
+ensure_network_online() {
+  if command -v ensure_network_online >/dev/null 2>&1; then
+    command ensure_network_online
+    return $?
+  fi
+  return 1
+}
+
+# --- timestamp helper ---
+nowstamp() { date +%Y%m%d%H%M%S 2>/dev/null || printf '%s\n' "now"; }
--- a/Runner/utils/lib_hwprobe.sh
+++ b/Runner/utils/lib_hwprobe.sh
@@ -1,0 +1,347 @@
+#!/bin/sh
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+# hw-probe helpers. Requires lib_common.sh, lib_apt.sh, lib_docker.sh
+
+HWPROBE_PKG="hw-probe"
+HWPROBE_DEPS="lshw smartmontools nvme-cli hdparm pciutils usbutils dmidecode ethtool lsscsi iproute2"
+# Flag for callers (e.g., run.sh) to know if we installed hw-probe in this run.
+# shellcheck disable=SC2034 # used by run.sh; not read within this file
+HWPROBE_INSTALLED_THIS_RUN=${HWPROBE_INSTALLED_THIS_RUN:-0}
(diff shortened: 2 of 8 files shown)


Runner/utils/lib_common.sh
```diff
@@ -0,0 +1,76 @@
...
+  d="$1"
+  if [ ! -d "$d" ]; then
+    mkdir -p "$d" 2>/dev/null || { need_root; $SUDO mkdir -p "$d" || return 1; }
+  fi
+}
+
+# --- OS detect (Debian/Ubuntu) ---
+detect_os_like() {
+  if [ -r /etc/os-release ]; then
+    # shellcheck source=/dev/null
```
lool, line 48: meh :)

Would this make shellcheck happy enough?
`(. /etc/os-release && echo "${ID_LIKE:-$ID}")`

Runner/utils/lib_hwprobe.sh
```diff
@@ -0,0 +1,171 @@
+#!/bin/sh
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+# hw-probe helpers. Requires lib_common.sh, lib_apt.sh, lib_docker.sh
+
+HWPROBE_PKG="hw-probe"
+HWPROBE_DEPS="lshw smartmontools nvme-cli hdparm pciutils usbutils dmidecode ethtool lsscsi iproute2"
```
lool, line 7: Looks like hdparm is a Depends of hw-probe already; I wonder how was this list generated? How would it be kept up-to-date?

Runner/utils/lib_hwprobe.sh
```diff
@@ -0,0 +1,171 @@
...
+  need_root
+  tmp="${out%/}/.hw-probe-run-$(nowstamp).log"
+  sh -c "$SUDO $cmd" 2>&1 | tee "$tmp"
+ 
+  # Extract URL if uploaded
+  url="$(sed -n 's/^.*Probe URL: *\([^ ]*linux-hardware\.org[^ ]*\).*$/\1/p' "$tmp" | tail -n 1)"
+  [ -n "$url" ] && log_info "Probe uploaded: $url"
+ 
+  # Parse saved artifact
+  saved="$(sed -n 's/^Saved to:[[:space:]]*//p' "$tmp" | tail -n 1)"
```
lool, line 82: This looks clean, but seems super complicated – parsing the output or log to figure where this was saved; is there no way to tell hw-probe to save somewhere specific? Or could we send a patch to do so? Seems painful to extract the info like this

Runner/utils/lib_hwprobe.sh
```diff
@@ -0,0 +1,171 @@
...
+  if [ "$upload" = "yes" ] && ! network_is_ok; then
+    log_warn "No network. Disabling upload for this Docker run."
+    upload="no"
+  fi
+
+  docker_install
+  docker_can_run || die "Docker installed but cannot run. Check permissions."
+
+  DCMD="$(docker_cmd)"
+  IMAGE="linuxhw/hw-probe"
```
lool, line 119: If we are using a docker image, shouldn't it contain all the probing tools we need already?

lool (commented):
Nice! So this is my first review here, I might have to ask some basic questions as I get familiar with the assumptions here

Here are my notes in no particular order:
1) the first line of the commit messages is a bit long; I guess it's just a draft and would be cleaned up, I see other commits even seem to use semantic git commits, nice!

2) a description of the "why" would be nice somewhere

3) the shell code is super clean, big kudos for the style!

4) I don't understand why we support installing the package or using the docker image; is it to compare? are there pros and cons to each?

5) I also don't understand why we support optional core utilities like tar and the like being absent; is this for portability to BSD or something? Seems a bit extreme to have to carry this deep in shell scripts

6) this is not really feedback on this specific test, but more on what I infer from the architecture: I like how the shell scripts are nicely independent in this repo, even if they use some common libraries; it looks like each script will pull its dependencies as it starts, which I find a bit weird. I would either expect:
- we build an image with everything in it, and then it gets used or tests are skipped
- or we embed the dependencies and don't touch the image
- or we dynamically install everything missing for what we want to run at the start, then run all tests; that way, we control internet accesses during the run, and any network / package install errors are flushed out early (downside: can be long to install all deps before seeing some early tests fail)

Perhaps this could be implemented with dependency injection: each test would provide a few functions – "can_run()", "list_missing_os_pkgs()", "run_tests()" etc. and the main testsuite would call the tests' functions to report their needs. Or the dependencies would be metadata next to the tests that get processed by the test itself or by the testsuite.

7) I find the optional upload feature nice, but I'm curious how we'd use it; would we setup separate jobs where it's enabled?

Again, apologies for my naive questions, this looks great!

Runner/utils/lib_hwprobe.sh
```diff
@@ -0,0 +1,171 @@
...
+}
+
+# Extract a .txz (xz-compressed tar) to an output dir, with fallbacks.
+# $1=saved_archive $2=dest_dir
+_hwprobe_extract_txz() {
+  saved="$1"; dest="$2"
+  [ -f "$saved" ] || { log_warn "Cannot extract: file not found: $saved"; return 1; }
+  ensure_dir "$dest" || { log_warn "Cannot create extract dir: $dest"; return 1; }
+
+  # Prefer tar -J; fall back to bsdtar; then xz | tar
```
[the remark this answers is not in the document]
smuppand: Since we target Debian/Ubuntu, GNU tar with -J is available. We can simplify and remove the bsdtar/xz fallbacks. We already check for tar in run.sh when --extract yes.

Runner/utils/lib_hwprobe.sh
smuppand: Yes. For Docker mode we don’t install any host tools—only ensure Docker itself exists. The linuxhw/hw-probe image bundles what it needs, we pass mounts/privileged flags so it can see host devices. Our current flow respects that (no host deps installation for Docker runs).

We’ll clarify this in the README (“Docker mode does not install host utilities, image contains them”).

smuppand:
> Nice! So this is my first review here, I might have to ask some basic questions as I get familiar with the assumptions here
> 
> Here are my notes in no particular order:
> 
> 1. the first line of the commit messages is a bit long; I guess it's just a draft and would be cleaned up, I see other commits even seem to use semantic git commits, nice!
> 2. a description of the "why" would be nice somewhere
> 3. the shell code is super clean, big kudos for the style!
> 4. I don't understand why we support installing the package or using the docker image; is it to compare? are there pros and cons to each?

Native (APT):
Pros: integrates with host, smallest runtime overhead, works without Docker.
Cons: needs root & network (first install), relies on host packages.

Docker:
Pros: self-contained userspace (image has tools), reproducible, minimal host mutation.
Cons: needs Docker, --privileged + host mounts, initial image pull.

> 5. I also don't understand why we support optional core utilities like tar and the like being absent; is this for portability to BSD or something? Seems a bit extreme to have to carry this deep in shell scripts

Totally fair. We target Debian/Ubuntu, so GNU tar with -J is effectively guaranteed. I’ve simplified extraction accordingly:

_hwprobe_extract_txz() now requires GNU tar (tar -xJf).

We also require tar up front when --extract yes is used.


> 6. this is not really feedback on this specific test, but more on what I infer from the architecture: I like how the shell scripts are nicely independent in this repo, even if they use some common libraries; it looks like each script will pull its dependencies as it starts, which I find a bit weird. I would either expect:
> 
> * we build an image with everything in it, and then it gets used or tests are skipped

This is the good idea and we are currently following this apporach in current qcom-linux-testkit as well. 

> * or we embed the dependencies and don't touch the image
> * or we dynamically install everything missing for what we want to run at the start, then run all tests; that way, we control internet accesses during the run, and any network / package install errors are flushed out early (downside: can be long to install all deps before seeing some early tests fail)
> 
> Perhaps this could be implemented with dependency injection: each test would provide a few functions – "can_run()", "list_missing_os_pkgs()", "run_tests()" etc. and the main testsuite would call the tests' functions to report their needs. Or the dependencies would be metadata next to the tests that get processed by the test itself or by the testsuite.
> 
> 7. I find the optional upload feature nice, but I'm curious how we'd use it; would we setup separate jobs where it's enabled?

We’ll expose it in CI via a matrix or separate job:

UPLOAD=no by default (save artifacts locally for later inspection).

UPLOAD=yes in an opt-in job (e.g., nightly / lab runs) to seed [linux-hardware.org](http://linux-hardware.org/) the runner prints the URL for traceability.


This keeps most CI jobs fast and isolated while allowing a dedicated pipeline to contribute probes.

> 
> Again, apologies for my naive questions, this looks great!

smuppand:
> I'm sorry this PR has expired, it's still very much on my nice to have list; I'm not sure where we left things for the code changes in this PR, happy to resume. However I do know I currently lack a way of running these tests in some CI of the Debian images. We could do it either with Axiom integration (I've been working on that but am currently stuck), or LAVA template wrapping of these tests (which I've seen implemented elsewhere, but I didn't have time to implement).

I will resume work on this after I finish my current tasks next week. Could you please review the current logic and share any missing requirements? I will address them once I return to this.

PR #170, opened 2025-09-18
fastrpc_test: export ASSETS_DIR for back-compat, silence ShellCheck SC2034

smuppand:
Why

- Our fastrpc_test runner moved to the /usr/local layout and no longer uses --assets-dir.
- ShellCheck now flags ASSETS_DIR as unused (SC2034).
- Some legacy wrappers may still pass --assets-dir and expect the variable to be visible to child processes.

What changed

- After argument parsing, if --assets-dir was provided, we now:
- export ASSETS_DIR
- log a single compat line: (compat) --assets-dir provided: … (ignored with /usr/local layout)
- No functional changes to execution, arguments, pass/fail logic, or environment setup beyond exporting the variable.

### DIFF
--- a/Runner/plans/fastrpc-premerge.yaml
+++ b/Runner/plans/fastrpc-premerge.yaml
@@ -14,6 +14,6 @@ metadata:
 run:
     steps:
         - cd Runner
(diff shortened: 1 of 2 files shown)


abbajaj806 (approved):
Looks Good to me

PR #182, opened 2025-10-06
IRIS Video: stabilize stack switching & avoid duplicate heavy ops in --stack both; cleaner logs and safer Kodiak FW flow

smuppand:
Introduces top-level vs sub-run gating in [run.sh](http://run.sh/) so heavy operations run once per --stack both execution:

- ensure_rootfs_min_size
- initial network bring-up (ensure_network_online)
- early media bundle fetch/extract

Adds robust post-stack capture in [run.sh](http://run.sh/):

- video_dump_stack_state before/after
- whitespace-safe normalization of post_stack with fallback to video_stack_status when unknown

Keeps Kodiak firmware handling safe and minimal:

- upstream FW install done once per pass
- downstream reload only when requested/needed, with existing [lib_video.sh](http://lib_video.sh/) helpers

- Preserves CLI and logic; no renames, no one-liners. Logging now explicitly states when sub-runs skip costly steps (e.g., “Sub-run: skipping rootfs size check”).

[lib_video.sh](http://lib_video.sh/) enhancements leveraged:

- retrying modprobe, runtime block/unblock, persistent blacklist handling, V4L node refresh, FW swap/reload helpers

- [functestlib.sh](http://functestlib.sh/) & YAML: no interface changes; existing helpers (check_dependencies, extract_tar_from_url, ensure_rootfs_min_size, scan_dmesg_errors) used as-is.

Verification

- Example --stack both run shows single early fetch, sub-runs skip heavy steps, clear pre/post module dumps, clean dmesg triage, and PASS for all tests in both passes.

Compatibility & Risk

- Backwards compatible flags and defaults.
- Writes to /etc/modprobe.d remain intentional and idempotent.
- Requires root for module ops (existing requirement).


abbajaj806 (approved):
LGTM

PR #183, opened 2025-10-07
IRIS runner: add default app/firmware fallbacks and path probing (no logic changes)

smuppand:
- Respect --app when provided; otherwise resolve iris_v4l2_test by probing: $PATH, then /usr/bin/iris_v4l2_test, then /data/vendor/iris_test_app/iris_v4l2_test.
- For Kodiak downstream/overlay runs, if --downstream-fw is not passed, automatically use /data/vendor/iris_test_app/firmware/vpu20_1v.mbn when present.
- Emit clear logs showing which app/firmware path was selected to aid triage.
- No renames, no behavioral logic changes to the runner flow; only path resolution was added.
- Backwards compatible: existing CLI overrides still win; upstream-only builds continue to work.


abbajaj806 (approved):
Looks good to me

PR #200, opened 2025-10-22
AudioRecord: PipeWire-first recording with smart fallback and capture helpers added

smuppand:
Why this PR?

- Use pw-record against PipeWire default when SRC_ID is empty; never skip prematurely.
- Accept non-trivial output (bytes > 1024) even if rc ≠ 0 to avoid unnecessary fallbacks.
- Add alsa_pick_virtual_pcm() to pick pipewire → pulse → default via `arecord -L`.
- Harden ALSA path: normalize "hw:0 1," → "hw:0,1" and auto-pick a valid capture device.
- Preserve existing logs/JUnit; resolves RB3 mic detection and prevents spurious retries.


abbajaj806 (approved):
LGTM

PR #220, opened 2025-11-28
Added UFS functional tests

jeevanandan-sandan:


Runner/suites/Kernel/Baseport/Storage/ufs_gear_lane/run.sh
```diff
@@ -0,0 +1,266 @@
...
+
+TESTNAME="ufs_gear_lane"
+test_path=$(find_test_case_by_name "$TESTNAME")
+cd "$test_path" || exit 1
+res_file="./$TESTNAME.res"
+
+log_info "--------------------------------------------------"
+log_info "------------- Starting $TESTNAME Test ------------"
+
+check_dependencies dd grep cut head tail udevadm sleep
```
smuppand: cut, head, tail, udevadm are not used anywhere in this script. You do use findmnt and awk, and rely on readlink (though that's almost always present)

Runner/suites/Kernel/Baseport/Storage/ufs_gear_lane/run.sh
```diff
@@ -0,0 +1,266 @@
...
+    exit 1
+fi
+log_info "Expected UFS Gear: HS_GEAR$EXPECTED_GEAR"
+
+# Start dd command in background and check gear/lane under load
+log_info "Starting I/O load test to verify UFS Gear and Lane under load..."
+tmpfile="/ufs_gear_lane.tmp"
+
+# Start dd in background
+dd if=/dev/zero of="$tmpfile" bs=3M count=1024 >/dev/null 2>&1 &
```
smuppand: of="$tmpfile" is  on / -> whichever filesystem backs. you even skip when UFS is rootfs (resolved_block == resolved_rootfs), meaning in the common safe case, you're explicitly not touching UFS at all.

smuppand (changes requested):
These comments apply to the other tests as well. Please also take care of the https://github.com/qualcomm-linux/qcom-linux-testkit/actions/runs/19762511972/job/56772745461.

Runner/suites/Kernel/Baseport/Storage/ufs_gear_check/run.sh
```diff
@@ -0,0 +1,266 @@
...
+for cfg in $OPTIONAL_CONFIGS; do
+    if ! check_kernel_config "$cfg" 2>/dev/null; then
+        log_info "[OPTIONAL] $cfg is not enabled"
+        missing_optional="$missing_optional $cfg"
+    fi
+done
+[ -n "$missing_optional" ] && log_info "Optional configs not present but continuing:$missing_optional"
+
+check_dt_nodes "/sys/bus/platform/devices/*ufs*" || {
+    echo "$TESTNAME SKIP" > "$res_file"
```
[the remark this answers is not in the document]
jeevanandan-sandan: added log_skip when UFS dt_nodes are missing.

smuppand (changes requested):

Runner/suites/Kernel/Baseport/Storage/ufs_gear_check/run.sh
```diff
@@ -0,0 +1,266 @@
...
+# Get UFS version and determine expected gear
+spec_version_value=$(cat "$SPEC_VERSION" 2>/dev/null)
+if [ -z "$spec_version_value" ]; then
+    log_fail "Failed to read specification version from $SPEC_VERSION"
+    echo "$TESTNAME FAIL" > "$res_file"
+    exit 1
+fi
+log_info "UFS Specification Version: $spec_version_value"
+
+EXPECTED_GEAR=$(get_ufs_gear "$spec_version_value")
```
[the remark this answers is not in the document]
jeevanandan-sandan: Skipping the test case if the ufs spec is unknown.

Runner/suites/Kernel/Baseport/Storage/ufs_clock_scaling/run.sh
```diff
@@ -0,0 +1,186 @@
...
+    log_fail "UFS clock is not gated"
+else
+    log_pass "UFS clock is gated to $UFS_MAX_FREQ"
+    UFS_CLOCK_GATED=1
+fi
+
+if [ "$UFS_CLOCK_SCALED" -ne 1 ] && [ "$UFS_CLOCK_GATED" -ne 1 ]; then
+    log_fail "UFS clock scaling & gating test failed"
+    echo "$TESTNAME FAIL" > "$res_file"
+    exit 1
```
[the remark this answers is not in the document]
jeevanandan-sandan: Changed all the exit codes to zero irrespective of test result.

Runner/utils/functestlib.sh
```diff
@@ -4293,3 +4293,59 @@ get_pid() {
...
+    delay="$4"
+    
+    # Default values if not provided
+    [ -z "$retries" ] && retries=5
+    [ -z "$delay" ] && delay=1
+    
+    attempt=1
+    
+    echo "Checking status of node '$node_name', expecting: '$expected_status'"
+    echo "Will retry up to $retries times with $delay second(s) delay between attempts"
```
smuppand: It's the same here as well.

Runner/utils/functestlib.sh
```diff
@@ -4293,3 +4293,59 @@ get_pid() {
...
+    
+    while [ "$attempt" -le "$retries" ]; do
+        echo "Attempt $attempt of $retries..."
```
smuppand: It's the same here as well.

Runner/utils/functestlib.sh
```diff
@@ -4293,3 +4293,59 @@ get_pid() {
...
+        
+        # Get node status by directly reading from the file with the node's name
+        status=$(cat "$node_name")
+        
+        if [ "$status" = "$expected_status" ]; then
+            echo "Success! Node '$node_name' has status: '$status'"
+            return 0
+        else
+            echo "Current status: '$status', waiting for: '$expected_status'"
```
smuppand: It's the same here as well.

Runner/utils/functestlib.sh
```diff
@@ -4293,3 +4293,59 @@ get_pid() {
...
+            
+            if [ "$attempt" -lt "$retries" ]; then
+                echo "Retrying in $delay second(s)..."
```
smuppand: It's the same here as well.

Runner/utils/functestlib.sh
```diff
@@ -4293,3 +4293,59 @@ get_pid() {
...
+    attempt=1
+    
+    echo "Checking status of node '$node_name', expecting: '$expected_status'"
+    echo "Will retry up to $retries times with $delay second(s) delay between attempts"
+    
+    while [ "$attempt" -le "$retries" ]; do
+        echo "Attempt $attempt of $retries..."
+        
+        # Get node status by directly reading from the file with the node's name
+        status=$(cat "$node_name")
```
smuppand: this is unguarded, If file disappers/ unreadable - cat prints to stderr and status becomes empty. Make reads safe: `status=$(cat "$node_name" 2>/dev/null | tr -d '\n' | tr -d '\r')`

Runner/utils/functestlib.sh
```diff
@@ -4293,3 +4293,59 @@ get_pid() {
         return 1
     fi
 }
+
+# Function to get the first matching node path
+get_dt_node_path() {
```
smuppand: Make it accept multiple args: get_dt_node_path "/a/*" "/b/*" and loop over "$@" with glob expansion explicitly, or Keep single pattern but implement safe glob expansion (and handle “no match” deterministically).

If you keep it as-is, at least rename it to something like first_existing_path_from_globs and update callers.

Runner/suites/Kernel/Baseport/Storage/ufs_write_booster/run.sh
```diff
@@ -0,0 +1,188 @@
...
+log_info "Detected UFS block: $block_dev"
+
+# Function to compare UFS specification versions
+# Returns 0 if version1 >= version2, 1 otherwise
+compare_ufs_version() {
+    version1="$1"
+    version2="$2"
+    
+    # Convert hex to decimal for comparison
+    ver1_dec=$((version1))
```
smuppand: This works only if the shell accepts hex in arithmetic expansion and the string is exactly acceptable (e.g., 0x0310). If it’s 0310 or has whitespace/newline, it breaks.

strip newline, ensure prefix, or parse with printf %d.

Runner/suites/Kernel/Baseport/Storage/ufs_clock_scaling/run.sh
```diff
@@ -0,0 +1,189 @@
...
+    log_skip "Failed to get UFS max clock frequency node path"
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+log_info "Found UFS max freq node: $UFS_MAX_FREQ_NODE"
+
+UFS_MAX_FREQ=$(cat "$UFS_MAX_FREQ_NODE" 2>/dev/null)
+if [ -z "$UFS_MAX_FREQ" ]; then
+    log_skip "Failed to read max frequency from $UFS_MAX_FREQ_NODE"
+    echo "$TESTNAME FAIL" > "$res_file"
```
smuppand: Failing to read max_freq currently leads to FAIL (good), but elsewhere similar read failures are treated as SKIP. Rule of thumb:
“Feature/sysfs node not present” → SKIP
“Node present but unreadable / invalid value” → FAIL (because kernel/userspace is broken)

Runner/suites/Kernel/Baseport/Storage/ufs_clock_scaling/run.sh
```diff
@@ -0,0 +1,189 @@
...
+if [ -z "$UFS_CLOCK_FREQ_NODE" ]; then
+    log_skip "Failed to get UFS clock frequency node path"
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+log_info "Found UFS clock freq node: $UFS_CLOCK_FREQ_NODE"
+
+# Check for UFS clock max freq node and assign to variable
+log_info "Checking for UFS max clock freq node..."
+if ! check_dt_nodes "/sys/devices/platform/soc@0/*ufs*/devfreq/*ufs*/max_freq"; then
```
smuppand: is too broad. Might match unreleated nodes or multiple controllers. Detect actual UFS  dir once and reuse. 
e.g., locate .../ufshcd* / .../ufs base and store as UFS_SYSFS.

Runner/suites/Kernel/Baseport/Storage/ufs_read_writes/run.sh
```diff
@@ -0,0 +1,147 @@
...
+    DD_WRITE="dd if=/dev/zero of=$tmpfile bs=1M count=64"
+fi
+
+if $DD_WRITE >/dev/null 2>&1 &&
+   dd if="$tmpfile" of=/dev/null bs=1M count=64 >/dev/null 2>&1; then
+    log_pass "UFS I/O stress test passed"
+    if command -v stat >/dev/null 2>&1; then
+        stat --format="[INFO] Size: %s bytes File: %n" "$tmpfile"
+    else
+        find "$tmpfile" -printf "[INFO] Size: %s bytes File: %p\n"
```
smuppand: find -printf is not POSIX and absent in busybox builds sometimes.
If stat isn’t available, just ls -l "$tmpfile" or wc -c < "$tmpfile".

smuppand (changes requested):
Centralize common UFS pre-checks
Every script repeats:

- mandatory/optional config checks
- DT node check
- detect_ufs_partition_block

This should be a reusable helper in [functestlib.sh](http://functestlib.sh/), e.g.:
ufs_common_precheck that returns SKIP reason and exports UFS_BLOCK_DEV / UFS_SYSFS_BASE.
Less duplication = fewer inconsistencies.

Runner/suites/Kernel/Baseport/Storage/ufs_write_booster/run.sh
jeevanandan-sandan: The spec version node returns the value in the format specified by JEDEC (E.g. 0x0310)

Runner/utils/functestlib.sh
```diff
@@ -4456,3 +4456,174 @@ get_pid() {
...
+    log_info "Detected UFS block: $block_dev"
+    
+    # Export block_dev for use by calling script
+    export UFS_BLOCK_DEV="$block_dev"
+
+    return 0
+}
+
+# Find the matching dt node
+get_dt_node_path() {
```
smuppand: expand via set -- $pattern with unquoted expansion (it will expand globs), then iterate.
 
Or require callers to pass unquoted patterns (not acceptable; will break static analysis and is fragile).

Runner/utils/functestlib.sh
```diff
@@ -4456,3 +4456,174 @@ get_pid() {
         return 1
     fi
 }
+
+# UFS precheck function - validates common UFS requirements
```
smuppand: You added:
get_dt_node_path
check_node_status
check_node_status_with_timeout
ufs_precheck_common
The repo already has extensive helpers (and you’ve standardized function names), this risks:
name collisions (same function name defined elsewhere)

If they don’t, at least namespace them: ufs_get_dt_node_path, ufs_check_node_status, etc., to avoid generic collisions.

smuppand (changes requested):

smuppand (changes requested):

smuppand (changes requested):
Now the repo license is changed to BSD-3-Clause. Fix it in all the files. 
scan_dmesg_errors argument order got changed. scan_dmesg_errors "$test_path" "ufs"

Runner/suites/Kernel/Baseport/Storage/ufs_clock_scaling/run.sh
```diff
@@ -0,0 +1,186 @@
...
+    SEARCH=$(dirname "$SEARCH")
+done
+
+if [ -z "$INIT_ENV" ]; then
+    echo "[ERROR] Could not find init_env (starting at $SCRIPT_DIR)" >&2
+    exit 0
+fi
+
+# Only source if not already loaded
+if [ -z "$__INIT_ENV_LOADED" ]; then
```
smuppand, line 24: The guard is inverted. This sources init_env only when __INIT_ENV_LOADED is already set. On a normal first-time run, init_env will not be sourced, but the script still immediately sources "$TOOLS/[functestlib.sh](http://functestlib.sh/)"

recommended fix, applies to other run.sh files as well
```
if [ -z "${__INIT_ENV_LOADED:-}" ]; then
    # shellcheck disable=SC1090
    . "$INIT_ENV"
fi
```

Runner/suites/Kernel/Baseport/Storage/ufs_clock_scaling/run.sh
```diff
@@ -0,0 +1,186 @@
...
+
+TESTNAME="ufs_clock_scaling"
+test_path=$(find_test_case_by_name "$TESTNAME")
+cd "$test_path" || exit 0
+res_file="./$TESTNAME.res"
+
+log_info "--------------------------------------------------"
+log_info "------------- Starting $TESTNAME Test ------------"
+
+check_dependencies dd sleep
```
smuppand, line 42: The script also uses cat, tr, kill, sync, and rm, but only checks dd sleep.

Runner/suites/Kernel/Baseport/Storage/ufs_clock_scaling/run.sh
```diff
@@ -0,0 +1,186 @@
...
+    exit 0
+fi
+log_info "Min UFS clock frequency supported is $UFS_MIN_FREQ"
+
+# Generate load using dd command and check for ufs cur_freq
+log_info "Starting I/O load test to verify UFS clock scaling..."
+tmpfile="$test_path/${TESTNAME}.tmp"
+
+# checking if necessary space is available in rootfs
+ensure_rootfs_min_size 3
```
smuppand, line 112: Return value is ignored. If rootfs is too small, the test still starts a 2 GiB write, which can fill the filesystem or create noisy CI failures.
Recommended fix:
```
if ! ensure_rootfs_min_size 3; then
    log_skip "Insufficient rootfs free space for UFS clock scaling load"
    echo "$TESTNAME SKIP" > "$res_file"
    exit 0
fi
```
This same fix applies to ufs_gear_check, ufs_hibern8, and ufs_write_booster.

Runner/suites/Kernel/Baseport/Storage/ufs_clock_scaling/run.sh
```diff
@@ -0,0 +1,189 @@
...
+    exit 0
+fi
+log_info "Max UFS clock frequency supported is $UFS_MAX_FREQ"
+
+# Generate load using dd command and check for ufs cur_freq
+log_info "Starting I/O load test to verify UFS clock scaling..."
+tmpfile="/ufs_clock_scaling.tmp"
+
+# Start dd in background
+dd if=/dev/zero of="$tmpfile" bs=2M count=1024 >/dev/null 2>&1 &
```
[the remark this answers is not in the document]
smuppand: This writes a 2 GiB tmpfile in the test directory. This is heavy for CI, can be slow on smaller devices, and can cause space pressure. The test only needs enough sustained write pressure to trigger devfreq it does not necessarily need a fixed 2 GiB write on every board.

Recommended fix: Parameterize the load size and default to a smaller value, for example:
```
UFS_LOAD_MB="${UFS_LOAD_MB:-512}"
count=$((UFS_LOAD_MB / 2))
dd if=/dev/zero of="$tmpfile" bs=2M count="$count" >/dev/null 2>&1 &
```
Set larger values from YAML/LAVA only when needed.

Runner/suites/Kernel/Baseport/Storage/ufs_clock_scaling/run.sh
```diff
@@ -0,0 +1,189 @@
...
+
+# Wait for dd to complete if it's still running
+if kill -0 $DD_PID 2>/dev/null; then
+    log_info "Waiting for I/O load test to complete..."
+    wait $DD_PID 2>/dev/null
+    sync
+fi
+
+# Clean up temp file
+rm -f "$tmpfile"
```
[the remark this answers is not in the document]
smuppand: The script starts both dd and a continuous sync loop, but there is no trap to clean them up if the script is interrupted. Background processes can be left running on interruption or unexpected exit.
In LAVA/CI, orphaned dd or sync loops can interfere with later tests.
Recommended fix:
```
cleanup() {
    [ -n "${DD_PID:-}" ] && kill "$DD_PID" 2>/dev/null || true
    [ -n "${sync_pid:-}" ] && kill "$sync_pid" 2>/dev/null || true
    [ -n "${tmpfile:-}" ] && rm -f "$tmpfile"
}
trap cleanup EXIT INT TERM
```

Runner/suites/Kernel/Baseport/Storage/ufs_gear_check/run.sh
```diff
@@ -0,0 +1,182 @@
...
+# Run common UFS prechecks
+if ! ufs_precheck_common; then
+    log_skip "Required UFS configurations are not available"
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+log_info "UFS prechecks successful"
+
+# Get UFS gear based on specification version
+get_ufs_gear() {
```
smuppand, line 53: Move get_ufs_gear() to [functestlib.sh](http://functestlib.sh/) as ufs_expected_hs_gear_from_spec() or similar, if future tests will reuse it.

Runner/suites/Kernel/Baseport/Storage/ufs_gear_check/run.sh
```diff
@@ -0,0 +1,182 @@
...
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+log_info "Expected UFS Gear: HS_GEAR$EXPECTED_GEAR"
+
+log_info "Starting I/O load test to verify UFS Gear under load..."
+tmpfile="$test_path/${TESTNAME}.tmp"
+
+# checking if necessary space is available in rootfs
+ensure_rootfs_min_size 3
```
smuppand, line 144: Return code ignored before starting the 2 GiB write. Same risk as clock scaling: insufficient space can cause noisy failures.

Recommended fix: Gate the load:
```
if ! ensure_rootfs_min_size 3; then
    log_skip "Insufficient rootfs free space for UFS gear load"
    echo "$TESTNAME SKIP" > "$res_file"
    exit 0
fi
```

Runner/suites/Kernel/Baseport/Storage/ufs_hibern8/run.sh
```diff
@@ -0,0 +1,123 @@
...
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+
+# Generate load using dd command and check for ACTIVE state
+log_info "Starting I/O load test to verify UFS link state transitions..."
+tmpfile="$test_path/${TESTNAME}.tmp"
+
+# checking if necessary space is available in rootfs
+ensure_rootfs_min_size 3
```
smuppand, line 77: Return value is ignored. The test may start a 2 GiB write even when the space prerequisite failed.
Recommended fix: Same guarded pattern

Runner/suites/Kernel/Baseport/Storage/ufs_write_booster/run.sh
```diff
@@ -0,0 +1,165 @@
...
+# Returns 0 if version1 >= version2, 1 otherwise
+compare_ufs_version() {
+    version1="$1"
+    version2="$2"
+    
+    # Convert hex to decimal for comparison
+    # shellcheck disable=SC3052
+    ver1_dec=$((16#$(printf '%s' "$version1" | sed 's/^0x//')))
+    # shellcheck disable=SC3052
+    ver2_dec=$((16#$(printf '%s' "$version2" | sed 's/^0x//')))
```
smuppand, line 62: This may work in some shells but fail in POSIX/minimal shells. The repo preference is POSIX/ShellCheck-clean without bashisms. Use awk or printf conversion in a portable way

Runner/suites/Kernel/Baseport/Storage/ufs_write_booster/run.sh
```diff
@@ -0,0 +1,165 @@
...
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+
+# Generate load using dd command and check Write Booster status
+log_info "Starting I/O load test to verify Write Booster activation..."
+tmpfile="$test_path/${TESTNAME}.tmp"
+
+# checking if necessary space is available in rootfs
+ensure_rootfs_min_size 3
```
smuppand, line 126: Return value is ignored.  The test may start a 2 GiB write even when the space prerequisite failed.
Recommended fix: Same guarded pattern

Runner/suites/Kernel/Baseport/Storage/ufs_read_writes/run.sh
```diff
@@ -0,0 +1,138 @@
...
+resolved_rootfs=$(readlink -f "$rootfs_dev" 2>/dev/null)
+
+if [ -n "$resolved_block" ] && [ -n "$resolved_rootfs" ] && [ "$resolved_block" = "$resolved_rootfs" ]; then
+    log_warn "Detected block ($resolved_block) is the root filesystem. Skipping read test."
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+
+log_info "Running basic read test on $block_dev (non-rootfs)..."
+if echo | dd of=/dev/null iflag=direct 2>/dev/null; then
```
smuppand, line 77: This checks whether dd accepts iflag=direct, but it does not test whether direct I/O works with the actual UFS block device. 
Recommended fix: log it as “dd supports iflag=direct” rather than assuming the device will. Current fallback is acceptable because the later real command gates PASS/FAIL.

Runner/suites/Kernel/Baseport/Storage/ufs_read_writes/run.sh
```diff
@@ -0,0 +1,138 @@
...
+fi
+
+log_info "Running I/O stress test (64MB read+write on tmpfile)..."
+tmpfile="$test_path/${TESTNAME}.tmp"
+
+if echo | dd of=/dev/null conv=fsync 2>/dev/null; then
+    DD_WRITE="dd if=/dev/zero of=$tmpfile bs=1M count=64 conv=fsync"
+else
+    log_warn "'conv=fsync' not supported by dd. Using basic write."
+    DD_WRITE="dd if=/dev/zero of=$tmpfile bs=1M count=64"
```
smuppand, line 100: The tmpfile path is unquoted inside the constructed command string. Avoid command strings and call dd directly in branches, or keep the path safe with positional execution. Since paths here are controlled, this is non-blocking.

Runner/utils/functestlib.sh
```diff
@@ -5351,3 +5351,291 @@ partition_log_block_devices() {
...
+# Exports the UFS Block Device
+# Returns 0 if mandatory ufs configurations are enabled else returns 1 
+ufs_precheck_common() {
+    MANDATORY_CONFIGS="CONFIG_SCSI_UFSHCD CONFIG_SCSI_UFS_QCOM"
+    OPTIONAL_CONFIGS="CONFIG_SCSI_UFSHCD_PLATFORM CONFIG_SCSI_UFSHCD_PCI CONFIG_SCSI_UFS_CDNS_PLATFORM CONFIG_SCSI_UFS_HISI CONFIG_SCSI_UFS_EXYNOS CONFIG_SCSI_UFS_ROCKCHIP CONFIG_SCSI_UFS_BSG"
+    
+    log_info "Checking mandatory kernel configs for UFS..."
+    if ! check_kernel_config "$MANDATORY_CONFIGS" 2>/dev/null; then
```
smuppand, line 5363: Redirecting check_kernel_config stderr hides useful diagnostic output. Also, because check_kernel_config already logs config status, suppressing part of its output can reduce triage quality.
CI logs may not clearly show exactly which mandatory config failed.

Recommended fix:
```
if ! check_kernel_config "$MANDATORY_CONFIGS"; then
or explicitly log the missing config after checking each one
```

Runner/utils/functestlib.sh
```diff
@@ -5351,3 +5351,291 @@ partition_log_block_devices() {
...
+        fi
+    done
+
+    IFS=$old_ifs
+    printf '%s' "$found_node"
+    return 1
+}
+
+# Function to check node status with retries by reading from a file
+ufs_check_node_status() {
```
smuppand, line 5442: This helper starts a background continuous sync loop every time it checks a node. This can create unnecessary I/O and make helper behavior surprising.

Recommended fix: Split into two helpers:
ufs_check_node_status() should only poll.
Callers that need sync pressure should start/stop sync explicitly or pass an option.

Runner/utils/functestlib.sh
```diff
@@ -5351,3 +5351,291 @@ partition_log_block_devices() {
...
+
+get_primary_ufs_controller() {
+    user_partition=$1
+
+    if [ -z "$user_partition" ]; then
+        log_error 'usage: get_primary_ufs_controller MOUNT_POINT\n' >&2
+        return 2
+    fi
+
+    dev=$(mount | awk -v mp="$user_partition" '
```
smuppand, line 5570: Parsing mount output is less robust than findmnt, and findmnt is already used in some scripts.
Recommended fix: 
```dev=$(findmnt -n -o SOURCE "$user_partition" 2>/dev/null) ```
with /proc/mounts fallback when findmnt is unavailable.

smuppand (changes requested):
Have proper commit message along with body about the changes.

PR #231, opened 2025-12-17
Fix: AudioPlayback no longer requires network when clips exist

### LINKED ISSUE #229: AudioPlayback requires networking
AudioPlayback test requires networking to download media files. It should be an option, but should not prevent the test from running. Please add a flag to run.sh script to prevent connecting to WiFi by default.

tmoida:
Resolves #229

Problem:
AudioPlayback test was forcing network connections even when audio files were already present locally, causing tests to fail in offline environments.

Solution:
- Add --enable-network-download flag (opt-in, default: disabled)
- Add audio_check_clips_available() to check files before network ops
- Implement smart network gating: only connect if files missing AND flag enabled
- Auto-enable network download when WiFi credentials provided via CLI


mwasilew (changes requested):
Please make the commit title more descriptive. The `resolves` part should not be in the title.

mwasilew (approved):

PR #235, opened 2025-12-18
Connectivity: Ethernet test fast-path + smarter DHCP handling for minimal builds

smuppand:
This PR improves the Ethernet validation flow to be less disruptive, less flaky, and easier to debug on both full userland images (NetworkManager/systemd-networkd) and minimal/kernel-only style images.
What changed
- Ethernet/[run.sh](http://run.sh/)
  - Adds a fast-path: if an interface already has Link detected=yes and a valid IPv4 (non 169.254.x.x), the test skips bring-up + DHCP and directly validates connectivity via ping.
  - Keeps existing behavior: if a network manager is active, the script does not run udhcpc and instead waits for IP.
  - Improves debug logs for triage (interface snapshot + ping output).
- [functestlib.sh](http://functestlib.sh/)
  - Updates try_dhcp_client_safe() to skip DHCP when a valid IP already exists.
  - If systemctl is available and NetworkManager/systemd-networkd is active, it waits for IP rather than fighting the manager.
  - Retains the safe udhcpc script flow for minimal images and adds extra logging for DHCP decisions/outcome.
- Why
  - Avoids re-running link bring-up/DHCP when the interface is already configured.
  - Prevents conflicts with network managers on full images.
  - Maintains usability on kernel/minimal environments where systemctl may not exist.
  - Better logging helps debug intermittent link/IP issues quickly in CI/LAVA.

### DIFF
--- a/Runner/suites/Connectivity/Ethernet/README.md
+++ b/Runner/suites/Connectivity/Ethernet/README.md
@@ -5,64 +5,154 @@ SPDX-License-Identifier: BSD-3-Clause-Clear
 
 ## Overview
 
-This test case validates the basic functionality of the Ethernet interface (`eth0`) on the device. It checks for:
+This test case validates Ethernet connectivity on the target device using the qcom-linux-testkit framework.
+
+Unlike the older `eth0`-only flow, the current test:
+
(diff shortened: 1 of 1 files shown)


vnarapar (approved):
LGTM

PR #280, opened 2026-02-08
Added dmesg errors to stdout

vnarapar:
Added stdout for the dmesg error prints for scan_dmesg_errors function

### DIFF
--- a/Runner/utils/functestlib.sh
+++ b/Runner/utils/functestlib.sh
@@ -3305,6 +3305,9 @@ scan_dmesg_errors() {
 
     if [ -s "$DMESG_ERRORS" ]; then
         log_info "dmesg scan: found non-benign module errors in $DMESG_ERRORS (history: $DMESG_HISTORY)"
+        while IFS= read -r line; do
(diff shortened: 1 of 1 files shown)


Runner/utils/functestlib.sh
```diff
@@ -3281,6 +3281,8 @@ scan_dmesg_errors() {
  
     if [ -s "$DMESG_ERRORS" ]; then
         log_info "dmesg scan found non benign module errors in $DMESG_ERRORS (history: $DMESG_HISTORY)"
+		log_info "Below are  the errors"
+		cat "$DMESG_HISTORY"
```
smuppand: prefix each line through log_info
This keeps the timestamps/prefixes consistent

if [ -s "$DMESG_ERRORS" ]; then
    log_info "dmesg scan found non-benign module errors in $DMESG_ERRORS (history: $DMESG_HISTORY)"
    log_info "Non-benign dmesg matches:"
    while IFS= read -r line; do
        log_info "[dmesg] $line"
    done < "$DMESG_ERRORS"
    return 0
fi

PR #298, opened 2026-02-15
add CMA (Contiguous Memory Allocator) enablement test script

vnarapar:
Introduce a CMA enablement/validation script that verifies configuration, initialization, runtime behavior, and statistics on Qualcomm platforms.

Key capabilities:
- Kernel config validation: CONFIG_CMA, CONFIG_DMA_CMA, CONFIG_CMA_DEBUG, CONFIG_CMA_DEBUGFS (mandatory), plus optional CMA sizing and areas.
- Memory stats collection: total, free, used, and usage% with size sanity checks (≥ 1 MB).
- DT configuration checks: CMA size/alignment and reserved-memory regions.
- Runtime verification: dmesg for CMA reservation/alloc/release messages, error/warning detection, and activity counts.
- Sysfs/Debugfs interfaces: enumerate `/sys/kernel/debug/cma` areas and report per-area stats; include relevant `/proc/vmstat` counters.
- Result artifacts: writes `cma.res` (PASS/FAIL) and logs detailed output to the console.

### DIFF
--- a/Runner/suites/Kernel/Baseport/cma/cma.yaml
+++ b/Runner/suites/Kernel/Baseport/cma/cma.yaml
@@ -1,0 +1,23 @@
+metadata:
+    format: Lava-Test Test Definition 1.0
+    name: cma
+    description: "CMA (Contiguous Memory Allocator) enablement and functionality validation"
+    maintainer:
+        - vnarapar@qti.qualcomm.com
+    os:
+        - openembedded
+    scope:
+        - functional
+    devices:
+        - rb3gen2
+        - qcs6490
+        - qcs8300
+        - qcs9100
+        - sa8775p
+
+run:
+    steps:
+        - REPO_PATH=$PWD
+        - cd Runner/suites/Kernel/Baseport/cma
+        - ./run.sh
+        - $REPO_PATH/Runner/utils/send-to-lava.sh cma.res
--- a/Runner/suites/Kernel/Baseport/cma/run.sh
+++ b/Runner/suites/Kernel/Baseport/cma/run.sh
@@ -1,0 +1,216 @@
+#!/bin/sh
+
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Robustly find and source init_env
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+INIT_ENV=""
+SEARCH="$SCRIPT_DIR"
+while [ "$SEARCH" != "/" ]; do
+    if [ -f "$SEARCH/init_env" ]; then
+        INIT_ENV="$SEARCH/init_env"
+        break
+    fi
+    SEARCH=$(dirname "$SEARCH")
+done
+
+if [ -z "$INIT_ENV" ]; then
+    echo "[ERROR] Could not find init_env (starting at $SCRIPT_DIR)" >&2
+    exit 1
+fi
+
+if [ -z "$__INIT_ENV_LOADED" ]; then
+    # shellcheck disable=SC1090
+    . "$INIT_ENV"
+fi
+
+# shellcheck disable=SC1090,SC1091
+. "$TOOLS/functestlib.sh"
+
+TESTNAME="cma"
+test_path=$(find_test_case_by_name "$TESTNAME")
+cd "$test_path" || exit 1
+res_file="./$TESTNAME.res"
+
+log_info "================================================================================"
+log_info "============ Starting $TESTNAME Testcase ======================================="
+log_info "================================================================================"
+
+pass=true
+
+log_info "=== CMA Kernel Configuration Validation ==="
+
+# Core CMA configs
+CORE_CMA_CONFIGS="CONFIG_CMA CONFIG_DMA_CMA CONFIG_CMA_DEBUG CONFIG_CMA_DEBUGFS"
+
+if ! check_kernel_config "$CORE_CMA_CONFIGS"; then
+    log_fail "CMA kernel configuration not enabled"
+    pass=false
+else
+    log_pass "CMA kernel configuration validated"
+fi
+
+OPTIONAL_CMA_CONFIGS="CONFIG_CMA_SIZE_MBYTES CONFIG_CMA_SIZE_SEL_MBYTES CONFIG_CMA_AREAS"
+
+log_info "Checking optional CMA configurations..."
+for cfg in $OPTIONAL_CMA_CONFIGS; do
+    if check_optional_config "$cfg" 2>/dev/null; then
+        log_pass "  $cfg: enabled"
+    else
+        log_warn "  $cfg: not enabled (optional)"
+    fi
+done
+
+if check_kernel_config "CONFIG_CMA_DEBUGFS" 2>/dev/null; then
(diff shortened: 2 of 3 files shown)


Runner/suites/Kernel/Baseport/cma/run.sh
```diff
@@ -0,0 +1,188 @@
...
+log_info "================================================================================"
+log_info "============ Starting $TESTNAME Testcase ======================================="
+log_info "================================================================================"
+
+pass=true
+
+log_info "=== CMA Kernel Configuration Validation ==="
+
+# Core CMA configs
+CORE_CMA_CONFIGS="CONFIG_CMA CONFIG_DMA_CMA CONFIG_CMA_DEBUG CONFIG_CMA_DEBUGFS"
```
smuppand: Right now you warn only. But you also fail earlier if CONFIG_CMA_DEBUGFS is missing. That’s okay for “enablement” coverage, but you should also emit a clear reason:

If config says enabled but path missing -> likely debugfs not mounted or permission issue -> WARN + remediation

If config not enabled ->  FAIL (enablement missing)


Extension (recommended):

call a helper like ensure_debugfs_mounted (if you have one), or implement minimal:

check mountpoint, try mount if root.

Runner/suites/Kernel/Baseport/cma/run.sh
```diff
@@ -0,0 +1,188 @@
...
+log_info "================================================================================"
+
+if $pass; then
+    log_pass "$TESTNAME : Test Passed"
+    echo "$TESTNAME PASS" > "$res_file"
+    exit 0
+else
+    log_fail "$TESTNAME : Test Failed"
+    echo "$TESTNAME FAIL" > "$res_file"
+    exit 1
```
smuppand: I’d recommend `exit 0` always (and rely on .res)

smuppand (changes requested):
Right now the test is mostly presence + observation. To validate CMA “functionality”, add one or more active functional checks:

A) Validate CMA is actually being used by at least one subsystem (non-invasive)
Check for common multimedia allocations that use CMA:
ION / dma-buf heaps (depending on kernel/userspace)
camera/video drivers that allocate contiguous buffers


Practical checks:
Look for dma_alloc_from_contiguous users in dmesg? (rare)
Check for increases in CmaFree drop after starting a known multimedia test (optional integration)

Extension idea: If media-ctl/camera nodes exist, do a short yavta capture (or your existing camera smoke test) and re-read /proc/meminfo before/after:

If CMA total is non-zero and capture runs, expect CmaFree to fluctuate at least once.

Don’t hard-fail if it doesn’t (drivers may use other paths), but log as INFO.

B) Validate reserved-memory DT looks correct (stronger DT parsing)
Current DT check only counts regions. You can improve by:
Identify nodes with compatible = "shared-dma-pool" and/or linux,cma-default
Report their size, alignment, reusable
Verify at least one CMA pool exists in DT when kernel cmdline doesn’t specify cma=


Implementation detail:
Read /proc/device-tree/reserved-memory/*/compatible, /linux,cma-default, /size, /alignment.
Size properties are binary cells; needs hexdump parsing (doable with od -An -t u4 etc).

C) Validate kernel cmdline CMA overrides
Check /proc/cmdline for:
cma= or cma_pernuma= etc (if relevant)
If present, compare configured size against /proc/meminfo CmaTotal with tolerance.

D) Validate CMA areas and per-area stats from debugfs (more than listing names)
In debugfs CMA directories, there are often files like:
count, pages, used, maxchunk etc (varies by kernel) Your script only lists area names. Enhance to dump key files if present:

total pages
free pages
used pages
alloc failures counters
Make it resilient: “if file exists, print it”.

E) Proactively detect CMA allocation failures in logs (more accurate)
Instead of generic error|fail|warn, look for common CMA failure strings:
cma_alloc: alloc failed
Failed to allocate contiguous memory
dma_alloc_from_contiguous: ... failed
cma: ... allocation failed
This reduces false positives from generic “warn”.

F) Add “pressure” functional test (optional, gated)
If you want a true kernel functional test without requiring multimedia:
Use cma_test module if present (some trees have it)
Or allocate contiguous memory via a small C helper using dma_heap (but that’s userspace and device-dependent)
Or run a workload that triggers hugepage / buddy allocator fragmentation and see if CMA still allocates (hard to do reliably)

Runner/suites/Kernel/Baseport/cma/run.sh
vnarapar: Updated

Runner/suites/Kernel/Baseport/cma/run.sh
```diff
@@ -0,0 +1,216 @@
...
+            fi
+        else
+            log_warn "CONFIG_CMA_DEBUGFS is enabled but /sys/kernel/debug/cma not found"
+        fi
+    fi
+fi
+
+log_info "=== CMA Memory Statistics ==="
+
+if [ -f "/proc/meminfo" ]; then
```
smuppand: This section reads CmaTotal/CmaFree, but can we also check /proc/cmdline for cma= overrides and compare the configured CMA size against CmaTotal with a reasonable tolerance? That would make the memory validation stronger than pure observation.

Runner/suites/Kernel/Baseport/cma/run.sh
```diff
@@ -0,0 +1,216 @@
...
+        pass=false
+    fi
+else
+    log_fail "/proc/meminfo not accessible"
+    pass=false
+fi
+
+if [ -d "/proc/device-tree/reserved-memory" ]; then
+    region_count=0
+    for region in /proc/device-tree/reserved-memory/*; do
```
smuppand: This DT validation is currently just a reserved-memory node count. Can we strengthen it by parsing the reserved-memory nodes for compatible = "shared-dma-pool", linux,cma-default, and log size/alignment/reusable when present? That would align the implementation better with the README claims.

Runner/suites/Kernel/Baseport/cma/cma.yaml
```diff
@@ -0,0 +1,21 @@
...
+        - functional
+    devices:
+        - rb3gen2
+        - qcs6490
+        - qcs8300
+        - qcs9100
+        - sa8775p
+
+run:
+    steps:
```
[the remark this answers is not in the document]
smuppand: > Updated

The current relative cd Runner/... form is less robust.

smuppand (changes requested):

smuppand (changes requested):

Runner/suites/Kernel/Baseport/cma/run.sh
```diff
@@ -0,0 +1,216 @@
...
+        fi
+    fi
+fi
+
+log_info "=== CMA Memory Statistics ==="
+
+if [ -f "/proc/meminfo" ]; then
+    if grep -q "CmaTotal" /proc/meminfo; then
+        cma_total=$(grep "CmaTotal" /proc/meminfo | awk '{print $2}')
+        cma_free=$(grep "CmaFree" /proc/meminfo | awk '{print $2}')
```
smuppand: Check /proc/cmdline for cma= overrides and compare the configured size against CmaTotal with a reasonable tolerance.

smuppand (changes requested):

PR #309, opened 2026-02-21
Removing Invalid secret, SEMGREP_APP_TOKEN is not defined in the reference of workflow

smuppand:
This secret is not used by the current Semgrep/preflight workflow and has been removed to avoid misconfiguration.


abbajaj806 (approved):
LGTM

PR #311, opened 2026-02-22
Harden DSP_AudioPD adsprpcd validation and add PID helper utilities

smuppand:
This PR improves the robustness of the Multimedia/DSP_AudioPD test and fixes the issue reported #313 , reusable PID utilities to [functestlib.sh](http://functestlib.sh/) to reduce CI/LAVA flakiness. 
 
Changes
- Runner/utils/[functestlib.sh](http://functestlib.sh/)
  - Add PID helper utilities:
    - sanitize_pid(): extract a single numeric PID from noisy/multi-line output
    - get_one_pid_by_name(): return the first matching PID for a process name
    - wait_pid_alive(): wait for a PID to become alive with a timeout
 
- Runner/suites/Multimedia/DSP_AudioPD/[run.sh](http://run.sh/)
  - Add log_info before dependency checks and SKIP when dependencies are missing
  - Use shared PID helpers for stable process discovery after daemon start/daemonize
  - Prefer /proc/<pid>/wchan and accept optimized symbol suffixes (e.g. poll_schedule_timeout.constprop.*)
  - Fall back to /proc/<pid>/stack; SKIP if neither interface is available
  - Kill adsprpcd only when started by the test

### DIFF
--- a/Runner/suites/Multimedia/DSP_AudioPD/run.sh
+++ b/Runner/suites/Multimedia/DSP_AudioPD/run.sh
@@ -2,6 +2,9 @@
 
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
 # SPDX-License-Identifier: BSD-3-Clause
+
+TESTNAME="DSP_AudioPD"
+
 # Robustly find and source init_env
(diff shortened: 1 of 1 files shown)


abbajaj806 (approved):
LGTM

PR #335, opened 2026-03-08
ci(shelllint): align PR linting with local ShellCheck runs

smuppand:
Summary
Update the shelllint workflow so pull requests lint only changed .sh files while reporting ShellCheck findings the same way they appear in local runs.

Changes

- keep PR lint scoped to changed shell files only
- remove the warning-only severity filter so style and info issues are also reported
- keep workflow steps POSIX sh basedpin ShellCheck installation to 0.9.0-1
- fetch enough git history for reliable PR diff detection

Result

- Shell lint results in GitHub Actions now match the manual shellcheck -s sh ... behavior more closely, while preserving the existing PR-only changed-file flow.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -11,29 +11,44 @@ jobs:
   shellcheck:
     runs-on: ubuntu-latest
 
+    defaults:
+      run:
+        shell: sh
+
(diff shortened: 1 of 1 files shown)


vnarapar (approved):
lgtm

PR #351, opened 2026-03-16
ci: align test inventory Pages workflow with Node 24 transition

smuppand:
This PR updates the test inventory GitHub Pages workflow to reduce the
current GitHub Actions runtime deprecation warnings.
 
Why:
- the workflow is publishing correctly, but Actions shows warnings for
  Node.js 20-based actions
- GitHub has announced the transition to Node.js 24 on hosted runners
- we should proactively align the workflow with that transition
 
What changed:
- updated actions/checkout from v4 to v6
- enabled FORCE_JAVASCRIPT_ACTIONS_TO_NODE24 at workflow level

### DIFF
--- a/.github/workflows/test-inventory-pages.yml
+++ b/.github/workflows/test-inventory-pages.yml
@@ -17,6 +17,9 @@ permissions:
   pages: write
   id-token: write
 
+env:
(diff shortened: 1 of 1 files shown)


abbajaj806 (approved):
LGTM

PR #360, opened 2026-03-22
shellcheck: fix trap-only cleanup false positives in kernel stress and userdata tests  

smuppand:
This PR fixes a small set of ShellCheck findings in two kernel test
scripts without changing runtime behavior.
 
What changed
- added a targeted SC2317 annotation for the trap-invoked cleanup
  function in:
  - Runner/suites/Kernel/Baseport/UserDataEncryption/[run.sh](http://run.sh/)
- updated Runner/suites/Kernel/Stress/Stressapptest/[run.sh](http://run.sh/) to:
  - add a targeted SC2317 annotation for the trap-invoked cleanup
    function
  - replace the CPU_COUNT calculation with a ShellCheck-clean form that
    preserves the intended counting of space-separated ONLINE_CPUS


vnarapar (approved):
LGTM

PR #306, opened 2026-02-20
Added QDSS-ETM scripts

Rohan-in-Qualcomm:
This PR introduces two new standalone test cases for the Embedded Trace Macrocell (ETM) drivers. These tests validate the ETM low-power mode toggling and the full trace data path enablement through standard sinks.

PR #370, opened 2026-03-26
Added QDSS-QPMDA / HW Events automated scripts

Rohan-in-Qualcomm:
This patch adds/updates a suite of automated scripts to validate the functionality of QDSS_QPMDA / HW Events.

* TPDM-Integration-Test: Verifies that TPDM nodes can be successfully linked to CoreSight sinks and that the overarching trace path is correctly established.
* TPDM-Trace-Sink: Validates end-to-end data generation and routing by enabling TPDMs, generating trace data, and verifying that the data can be successfully extracted from the /dev/ node of the targeted sink.
* utils/coresight_helper.sh: Common helper functions.


smuppand (changes requested):
This QDSS/QPMDA PR overlaps with the broader CoreSight helper stack.
 
Before merging this independently, I think we should align it with #389 so TPDM/QPMDA tests reuse the same shared helper patterns for debugfs discovery, source/sink discovery, enable/disable, trace extraction, cleanup, and SKIP handling.
 
If QDSS/QPMDA needs extra helper functions, those should either extend the common CoreSight helper or live in a clearly layered QDSS helper that builds on top of it. This will avoid duplicated tracing setup across #306/#370/#371/#389/#392/#393/#394.

PR #371, opened 2026-03-27
Added TGU Enable/Disable Validation Script

Rohan-in-Qualcomm:
* TGU-Enable-Disable: This script validates the state transtions of Coresight TGU devices. It dynamically discovers all the nodes and tries to toggle them on (1) or off (0) and checks for errors.

### DIFF
--- a/Runner/suites/Kernel/DEBUG/TGU-Enable-Disable/README.md
+++ b/Runner/suites/Kernel/DEBUG/TGU-Enable-Disable/README.md
@@ -1,0 +1,81 @@
+# Coresight TGU Enable/Disable Test
+
+## Overview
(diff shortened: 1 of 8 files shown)


smuppand (changes requested):
The TGU enable/disable validation should align with the common CoreSight helper direction from #389.
 
Please rebase after #389 is finalized and reuse the shared helper where possible for node discovery, state read/write, cleanup, and result reporting.
 
Since this test toggles hardware/debug nodes, please add robust cleanup traps and restore original state on EXIT/INT/TERM. Missing TGU/CoreSight/debugfs nodes should SKIP cleanly rather than FAIL on unsupported platforms.

PR #389, opened 2026-04-07
added coresight helper(utils) and sink-source-test

Rohan-in-Qualcomm:
- Coresight-Sink-Source-Test: Validate trace capture across all CoreSight source and sink combinations by enabling sources, routing to sinks, and verifying collected trace data.

Also a common library helper function coresight_helper.sh is added to prevent duplication of code blocks.

### DIFF
--- a/Runner/suites/Kernel/DEBUG/Coresight-Sink-Source-Test/Coresight-Sink-Source-Test.yaml
+++ b/Runner/suites/Kernel/DEBUG/Coresight-Sink-Source-Test/Coresight-Sink-Source-Test.yaml
@@ -1,0 +1,16 @@
+metadata:
+  name: Coresight-Sink-Source-Test
+  format: "Lava-Test Test Definition 1.0"
+  description: "This test iterates through all CoreSight sources and sinks to validate end‑to‑end trace path accessibility. It ensures trace data capture works correctly and that all sources are cleanly disabled after execution."
+  os:
+    - linux
+  scope:
+    - coresight
+    - kernel
+
+run:
+  steps:
+    - REPO_PATH=$PWD || true
+    - cd Runner/suites/Kernel/DEBUG/Coresight-Sink-Source-Test || true
+    - ./run.sh || true
+    - $REPO_PATH/Runner/utils/send-to-lava.sh Coresight-Sink-Source-Test.res || true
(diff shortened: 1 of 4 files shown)


Runner/suites/Kernel/DEBUG/Single-Sink-Reset-Connected-Base/run.sh
```diff
@@ -0,0 +1,166 @@
...
+    log_warn "Coresight directory $cs_base not found. Skipping test."
+    echo "0" > "$res_file"
+    exit 0
+fi
+
+stm=""
+for node in "$cs_base"/stm*; do
+    if [ -d "$node" ] && [ -f "$node/enable_source" ]; then
+        stm="$node"
+        brea
```
vnarapar: Can u check this?

Rohan-in-Qualcomm: Fixed it

smuppand (changes requested):
Thanks for adding the CoreSight helper and the sink/source validation test.
 
Since there are multiple pending CoreSight/QDSS PRs in parallel (#306, #370, #371, #392, #393, #394), I think we should treat #389 as the base/helper PR and review the rest as a coherent helper-based series.
 
Request:
1. Please make #389 establish the common CoreSight helper API that other tests can reuse.
2. Common operations such as debugfs/tracing path discovery, source/sink discovery, enable/disable, trace cleanup, sink reset, state restore, and result logging should live in the shared helper instead of being duplicated in each test.
3. Each test should have cleanup traps for EXIT/INT/TERM so enabled sources/sinks are disabled and trace paths are restored even if the test fails or is interrupted.
4. Missing CoreSight/debugfs/trace nodes should result in SKIP, not FAIL, unless the test explicitly targets a mandatory platform feature.
5. Please ensure scripts follow the existing repo conventions: init_env, [functestlib.sh](http://functestlib.sh/) logging, check_dependencies, .res generation, POSIX shell, ShellCheck clean, and LAVA-friendly exit behavior.
6. Please include LAVA/manual validation logs for at least one supported target showing the helper and sink/source test working end to end.
 
Once this helper API is agreed, the dependent PRs (#392/#393/#394) and the related QDSS/QPMDA/TGU/ETM PRs (#370/#371/#306) should rebase on top of it and reuse the helper instead of carrying separate discovery/setup/cleanup logic.

PR #392, opened 2026-04-08
Added coresight ftrace-dump-etf-base script

Rohan-in-Qualcomm:
- Ftrace-Dump-ETF-Base: Verify ftrace-to-STM linkage by capturing scheduler trace events through an ETF sink and validating trace output size.


smuppand (changes requested):
This PR depends on the CoreSight helper work in #389, so I suggest holding this until #389 is finalized.
 
After #389 lands, please rebase this PR and make sure Ftrace-Dump-ETF-Base uses the shared CoreSight helper for debugfs path discovery, source/sink setup, trace enable/disable, cleanup, and result handling.
 
The main thing I want to avoid is each CoreSight test carrying its own slightly different tracing setup and cleanup logic.

PR #393, opened 2026-04-08
Added coresight reset all and random sinks scripts

Rohan-in-Qualcomm:
- Reset-All-Sinks-Base: Stress test simultaneous reset of multiple CoreSight sinks and ensure all sinks reliably disable and recover for trace capture.

- Reset-Random-Sinks-Base: Randomly enable and reset multiple CoreSight sinks under stress and validate sink disable behavior and post-reset trace functionality.

### DIFF
--- a/Runner/suites/Kernel/DEBUG/Reset-All-Sinks-Base/README.md
+++ b/Runner/suites/Kernel/DEBUG/Reset-All-Sinks-Base/README.md
@@ -1,0 +1,87 @@
+# Reset-All-Sinks-Base Test
+
+## Overview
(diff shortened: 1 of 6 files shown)


smuppand (changes requested):
This reset-all/random-sinks test should be reviewed after #389 because it depends on the common CoreSight helper direction.
 
Please rebase on top of #389 once the helper API is finalized and reuse the shared helper for sink discovery, enable/disable, reset, trace cleanup, and state restore.
 
Since this test randomly toggles/reset sinks, cleanup is especially important. Please ensure EXIT/INT/TERM traps restore all enabled CoreSight state so later tests are not affected.

PR #394, opened 2026-04-08
Added single sink reset scripts

Rohan-in-Qualcomm:
- Single-Sink-Reset-Base: Repeatedly enable an active source and a single sink, reset CoreSight, and verify the sink consistently disables under load.

- Single-Sink-Reset-Connected-Base: Stress test sink reset behavior with an actively connected STM source and verify recovery via successful trace read.

### DIFF
--- a/Runner/suites/Kernel/DEBUG/Single-Sink-Reset-Base/README.md
+++ b/Runner/suites/Kernel/DEBUG/Single-Sink-Reset-Base/README.md
@@ -1,0 +1,93 @@
+# Single-Sink-Reset-Base
+
+## Overview
+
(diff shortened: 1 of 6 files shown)


smuppand (changes requested):
This PR depends on #389, so I suggest keeping it blocked until the common CoreSight helper API is finalized.
 
After #389 lands, please rebase this PR and reuse the shared helper for source/sink setup, single-sink reset, trace readback, cleanup, and result reporting.
 
The single-sink reset coverage is useful, but it should share the same discovery and cleanup path as the rest of the CoreSight tests.

PR #396, opened 2026-04-09
Fix AudioPlayback testcase ID collision for LAVA CI

tmoida:
## Description

This PR fixes the testcase ID collision issue in LAVA by implementing flexible testcase naming for the AudioPlayback test suite, enabling parallel execution of multiple AudioPlayback configurations in CI.

## Problem

LAVA was experiencing testcase ID collisions because:

- The AudioPlayback script was hardcoding the testcase name as `AudioPlayback` in the `.res` file
- When running multiple AudioPlayback configurations in parallel (e.g., Config01, Config03, Config05), all tests reported the same testcase ID
- LAVA expected unique testcase IDs like `AudioPlayback_Config01`, `AudioPlayback_Config03`, etc. based on the job definition
- This collision caused LAVA to overwrite test results and prevented proper CI enablement

## Solution

Implemented a flexible testcase naming mechanism that allows LAVA to specify unique testcase IDs:

- **Added `RESULT_TESTNAME` variable** - Defaults to `AudioPlayback` but can be overridden via CLI parameter
- **Added `--lava-testcase-id` parameter** - Allows LAVA to specify unique testcase names for each configuration
- **Added `LAVA_TESTCASE_ID` YAML parameter** - Passes the testcase ID from LAVA job definition to the script
- **Updated all result handling** - Changed all `.res` file writes to use `$RESULT_TESTNAME` instead of `$TESTNAME`
- **Pre-parse support** - Early parsing of `--lava-testcase-id` ensures unique result files even if setup fails

### DIFF
--- a/Runner/suites/Multimedia/Audio/AudioPlayback/AudioPlayback.yaml
+++ b/Runner/suites/Multimedia/Audio/AudioPlayback/AudioPlayback.yaml
@@ -27,10 +27,11 @@ params:
   NET_PROBE_ROUTE_IP: "1.1.1.1"  # IP used for route probing, default: 1.1.1.1
   NET_PING_HOST: "8.8.8.8"  # Host used for ping reachability check, default: 8.8.8.8
   RES_SUFFIX: ""  # Suffix for unique result file and log directory (e.g., "Config1" generates AudioPlayback_Config1.res and results/AudioPlayback_Config1/), default: unset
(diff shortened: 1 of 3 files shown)


tmoida:
Here are the details of the validation job:
https://lava.infra.foundries.io/scheduler/job/174081

smuppand (approved):
LGTM

PR #400, opened 2026-04-13
Reduce false BT and shmbridge failures caused by kernel log flooding

smuppand:
This PR also tightens the Bluetooth DT compatible matching so the BT test passes individual compatible strings instead of a single multiline pattern, avoiding misleading DT match output.
 
Summary of changes:
- use current-boot kernel log in btfwloaded()
- downgrade missing retained BT firmware-load signatures to WARN when runtime BT state is healthy
- replace shmbridge raw dmesg presence checks with sysfs/current-boot kernel log validation
- tighten BT DT compatible matching output
 
These changes are intended to improve robustness in LAVA and other noisy boot environments without weakening genuine failure coverage.


vnarapar (approved):
LGTM

PR #401, opened 2026-04-14
Fail weston-simple-egl on base when Weston runtime is missing

smuppand:
This PR fixes weston-simple-egl result handling on base builds so real Weston runtime regressions are reported as FAIL instead of SKIP.

This change updates the base-build result handling so these cases are treated as FAIL:
- connected display is present but no Wayland socket is found on base
- no Wayland socket is found after autodetection/retry on base
- Wayland connection test fails after retry

Why this change is needed:
A Weston upstream or integration regression on base can currently be masked as SKIP, which allows the change to pass post-merge validation without surfacing the failure. With this update, missing Weston runtime on base is treated as a real testcase failure, which is the expected behavior.
 
Expected outcome:
- base build + connected display + missing Weston runtime => FAIL
- base build + connected display + broken Wayland connection => FAIL
- no display / missing prerequisite => SKIP (unchanged)

### DIFF
--- a/Runner/suites/Multimedia/Graphics/weston-simple-egl/run.sh
+++ b/Runner/suites/Multimedia/Graphics/weston-simple-egl/run.sh
@@ -207,8 +207,8 @@ if [ -z "$sock" ]; then
       log_warn "overlay_start_weston_drm returned non-zero; private Weston may have failed to start."
     fi
   else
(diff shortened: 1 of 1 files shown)


bhargav0610 (approved):

bhargav0610:
LGTM

PR #415, opened 2026-04-22
ci: fix ShellCheck PR workflow merge-base resolution

smuppand:
Fix the PR ShellCheck workflow so it no longer fails before linting starts.
 
Changes
 
- use full checkout history
- fetch the base branch with an explicit remote-tracking refspec
- compute merge-base against `refs/remotes/origin/<base-branch>`
 
Why
 
The previous workflow could fail at `git merge-base` because `origin/<base>` was not reliably populated by the fetch step. That caused the job to exit before ShellCheck ran, so no lint issues were displayed.
 
Result
 
ShellCheck now runs correctly on changed `.sh` files in pull requests.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -33,9 +33,9 @@ jobs:
           set -eu
 
           BASE_BRANCH="${{ github.base_ref }}"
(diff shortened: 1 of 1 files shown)


abbajaj806 (approved):
LGTM

PR #435, opened 2026-05-01
add selectable Camera_NHX configs for CAMX validation

smuppand:
This PR extends Camera_NHX so CAMX validation can run selected NHX scenarios instead of only the default preview flow.
 
It adds target-specific NHX JSON configs for Kodiak, Lemans, Monaco, and Talos, adds --json/--target support in Camera_NHX, stages selected configs under /etc/camera/test/NHX as required by [nhx.sh](http://nhx.sh/), and updates CAMX DT matching to recognize current camera overlay compatibles.
 
Default behavior remains unchanged when no JSON is passed, while LAVA/manual runs can now explicitly validate preview, video, preview+video, and snapshot flows.


bhargav0610 (approved):
LGTM

PR #436, opened 2026-05-01
test: temporarily disable OpenCV in premerge CI

smuppand:
Temporarily disable OpenCV execution in the qcom-next premerge plan a
 
Why this is needed:
- The OpenCV standalone YAML path and the direct premerge [run.sh](http://run.sh/) path are not  currently equivalent.
- OpenCV.yaml was passing boolean flags such as --list and --shuffle as   key/value arguments, which can accidentally enable list-only behavior instead   of running the real tests.
- The premerge plan calls OpenCV/[run.sh](http://run.sh/) directly, so it executes the actual  default suite and can fail while the YAML path appears to pass.
- To avoid unstable or misleading premerge failures, OpenCV is temporarily disabled


vnarapar (approved):
LGTM

PR #438, opened 2026-05-03
kernel: strengthen smmu validation for QLI platforms

smuppand:
This PR extends the SMMU testcase under Runner/suites/Kernel/Baseport/smmu.
 
Why this PR is needed:
The existing smmu test only checked whether `dmesg | grep iommu` returned any output. That is too weak for QLI platforms and can both miss real issues and misrepresent actual SMMU status. On QLI systems, `/sys/kernel/debug/iommu` may be absent even when SMMU is functioning correctly, so relying on upstream-style
debugfs expectations is not sufficient.
 
What this PR changes:
- replaces the basic dmesg grep logic with stronger runtime validation
- checks required SMMU and IOMMU kernel configs
- validates `/sys/kernel/iommu_groups` presence and group enumeration
- validates platform devices attached to IOMMU groups
- checks critical DMA masters such as GPU, display, camera, video, UFS, USB,   ethernet, and audio, when present, for IOMMU protection
- uses shared functestlib helpers instead of raw one-off log parsing
- uses `scan_dmesg_errors` and `get_kernel_log` for SMMU or IOMMU related log analysis instead of a plain `dmesg | grep iommu`
 
Why this is useful:
On QLI hardware, the most meaningful SMMU validation is not just whether a log line exists, but whether:
- IOMMU groups are created
- DMA-capable devices are actually attached to those groups
- critical masters are not silently running without protection
- kernel logs do not show bypass, passthrough, or fault-related issues
 
This makes the testcase much closer to real SMMU runtime validation on target systems and helps catch regressions that the previous test would not detect.

### DIFF
--- a/Runner/suites/Kernel/Baseport/smmu/run.sh
+++ b/Runner/suites/Kernel/Baseport/smmu/run.sh
@@ -2,10 +2,14 @@
 
 # Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
 # SPDX-License-Identifier: BSD-3-Clause
-# Robustly find and source init_env
-SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+
+SCRIPT_DIR="$(
+    cd "$(dirname "$0")" || exit 1
+    pwd
+)"
 INIT_ENV=""
 SEARCH="$SCRIPT_DIR"
+
 while [ "$SEARCH" != "/" ]; do
     if [ -f "$SEARCH/init_env" ]; then
         INIT_ENV="$SEARCH/init_env"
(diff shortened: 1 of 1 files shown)


Runner/suites/Kernel/Baseport/smmu/run.sh
```diff
@@ -15,41 +19,231 @@ while [ "$SEARCH" != "/" ]; do
...
-# shellcheck disable=SC1090,SC1091
+
+# shellcheck disable=SC1091
 . "$TOOLS/functestlib.sh"
 
 TESTNAME="smmu"
-test_path=$(find_test_case_by_name "$TESTNAME")
+test_path="$(find_test_case_by_name "$TESTNAME")" || {
+    log_fail "$TESTNAME, test directory not found"
+    echo "$TESTNAME FAIL" > "./$TESTNAME.res"
```
vnarapar: is this check required?if u added for the lava/axiom framework to pick the status, i dont think itwill still fetch the result as fail as the .res path wont be found by test framework

Runner/suites/Kernel/Baseport/smmu/run.sh
```diff
@@ -15,41 +19,231 @@ while [ "$SEARCH" != "/" ]; do
...
 else
-    log_pass "$TESTNAME : Test Passed"
-    log_info "Writing to $res_file"
-    echo "$TESTNAME PASS" > "$res_file"
+    log_warn "CONFIG_ARM_SMMU_V3 is not enabled, continuing because some targets may use a different SMMU configuration"
+    inc_warn
+fi
+
+log_info "Checking IOMMU groups path"
+if command -v wait_for_path >/dev/null 2>&1; then
```
vnarapar: can this condition be removed? in else condition wait_for_path is again implemented which would nullify the advantage of defining a function

Runner/suites/Kernel/Baseport/smmu/run.sh
smuppand: > is this check required?if u added for the lava/axiom framework to pick the status, i dont think itwill still fetch the result as fail as the .res path wont be found by test framework

Agree. Removed the early .res write and fail early if the testcase path cannot be resolved.

Runner/suites/Kernel/Baseport/smmu/run.sh
smuppand: > can this condition be removed? in else condition wait_for_path is again implemented which would nullify the advantage of defining a function

functestlib.sh is already sourced and wait_for_path is available from shared library, the fallback condition is redundant. Removed the conditional fallback and call wait_for_path directly

vnarapar (approved):
LGTM

PR #439, opened 2026-05-03
utils: validate overlay audio setup without chmod or restart

### LINKED ISSUE #433: Audio test case steps update.
Hi Team,

We got merged change (https://github.com/qualcomm-linux/meta-qcom-distro/pull/230 )on meta-qcom for permission issue. now no need to chmod and restart the pipewire service.
Could please update the test steps?

Thanks,
Sairam

smuppand:
This PR updates the shared audio helper to align AudioPlayback and AudioRecord with the latest distro-side overlay audio fixes.
 
Fixes: #433
 
Why this PR is needed:
Issue #433 notes that meta-qcom-distro now handles the required dma_heap permission fix, so the audio tests no longer need to chmod /dev/dma_heap/system or restart PipeWire as part of the mandatory setup flow.
 
What this PR changes:
- updates Runner/utils/[audio_common.sh](http://audio_common.sh/)
- changes setup_overlay_audio_environment() from a mutating setup helper into a  non-mutating prerequisite validator
- removes the mandatory chmod of /dev/dma_heap/system from the overlay setup path
- removes the mandatory PipeWire restart from the overlay setup path
- validates that /dev/dma_heap/system is already accessible
- waits for PipeWire readiness without restarting it during setup

### DIFF
--- a/Runner/utils/audio_common.sh
+++ b/Runner/utils/audio_common.sh
@@ -496,74 +496,66 @@ audio_restart_pipewire_service() {
 }
 
 # Function: setup_overlay_audio_environment
-# Purpose: Configure audio environment for overlay builds (audioreach-based)
-# Returns: 0 on success, 1 on failure
-# Usage: Call early in audio test initialization, before backend detection
-# Configure overlay-specific audio prerequisites and restart PipeWire.
-# Restart and readiness share the same centralized timeout to avoid drift.
+# Purpose: Validate overlay audio prerequisites without mutating system state.
+# Returns: 0 on success, 1 on prerequisite failure
+# Usage: Call early in audio test initialization, before backend detection.
+#
+# Distro is expected to provide correct dma_heap permissions and PipeWire
+# readiness. This helper intentionally does not chmod /dev/dma_heap/system
+# and does not restart PipeWire, so distro regressions are not hidden by tests.
 setup_overlay_audio_environment() {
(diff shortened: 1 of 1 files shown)


Runner/utils/audio_common.sh
```diff
@@ -496,74 +496,56 @@ audio_restart_pipewire_service() {
...
-    log_info "Base build detected (no audioreach modules), skipping overlay setup"
+    log_info "Base build detected, no audioreach modules, skipping overlay setup"
     return 0
   fi
-
-  log_info "Overlay build detected (audioreach modules present), configuring environment..."
-
-  if [ "$(id -u)" -ne 0 ]; then
-    log_fail "Overlay audio setup requires root permissions"
+
```
abbajaj806, line 523: If lsmod itself fails (e.g., not available, permission denied), the pipe still exits non-zero — indistinguishable from "module not found". This silently skips overlay setup on a broken system.

smuppand: > If lsmod itself fails (e.g., not available, permission denied), the pipe still exits non-zero — indistinguishable from "module not found". This silently skips overlay setup on a broken system.

Now the helper fails clearly if lsmod is unavailable or if lsmod execution fails. It only skips overlay validation when lsmod succeeds and no audioreach module is present.

abbajaj806 (approved):
LGTM

PR #447, opened 2026-05-08
pelt scheduler: add new tests to validate pelt

vnarapar:
This PR adds new testcases to validate PELT scheduler
- This validates configs, decay, load tracking, schedutil and different tunables

### DIFF
--- a/Runner/suites/Kernel/Scheduler/PELT_config/PELT_config.yaml
+++ b/Runner/suites/Kernel/Scheduler/PELT_config/PELT_config.yaml
@@ -1,0 +1,23 @@
+metadata:
+  name: pelt-config
+  format: "Lava-Test Test Definition 1.0"
+  description: "Validates kernel configuration required for PELT (Per-Entity Load Tracking) scheduler"
+  maintainer:
+    - vnarapar@qti.qualcomm.com
+  os:
+    - linux
+  scope:
+    - functional
+  devices:
+    - rb3gen2
+    - qcs6490
+    - qcs8300
+    - qcs9100
+    - sa8775p
+
+run:
+  steps:
+    - REPO_PATH=$PWD
+    - cd "$REPO_PATH/Runner/suites/Kernel/Scheduler/PELT_config"
+    - ./run.sh
+    - $REPO_PATH/Runner/utils/send-to-lava.sh PELT_config.res
--- a/Runner/suites/Kernel/Scheduler/PELT_config/run.sh
+++ b/Runner/suites/Kernel/Scheduler/PELT_config/run.sh
@@ -1,0 +1,85 @@
+#!/bin/sh
+
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Robustly find and source init_env
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+INIT_ENV=""
+SEARCH="$SCRIPT_DIR"
+while [ "$SEARCH" != "/" ]; do
+    if [ -f "$SEARCH/init_env" ]; then
+        INIT_ENV="$SEARCH/init_env"
+        break
+    fi
+    SEARCH=$(dirname "$SEARCH")
+done
+
+if [ -z "$INIT_ENV" ]; then
+    echo "[ERROR] Could not find init_env (starting at $SCRIPT_DIR)" >&2
+    exit 1
+fi
+
+if [ -z "${__INIT_ENV_LOADED:-}" ]; then
+    # shellcheck disable=SC1090
+    . "$INIT_ENV"
+fi
+
+# shellcheck disable=SC1090,SC1091
+. "$TOOLS/functestlib.sh"
+
+TESTNAME="PELT_config"
+test_path=$(find_test_case_by_name "$TESTNAME")
+cd "$test_path" || exit 1
+res_file="./$TESTNAME.res"
+
+log_info "================================================================================"
+log_info "============ Starting $TESTNAME Testcase ======================================="
+log_info "================================================================================"
+log_info "Validates kernel configuration required for PELT (Per-Entity Load Tracking)"
+
+if ! check_dependencies zgrep; then
+    log_skip "$TESTNAME SKIP - missing required zgrep utility"
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+
+pass="true"
+
+log_info "=== Core PELT / CFS Kernel Configs ==="
+
+if [ ! -f /proc/config.gz ]; then
+    log_warn "/proc/config.gz not found  -  kernel config checks will be skipped"
+    log_warn "Ensure CONFIG_IKCONFIG and CONFIG_IKCONFIG_PROC are enabled in the kernel"
+	echo "$TESTNAME SKIP" > "$res_file"
+	exit 0
+else
+    CORE_CONFIGS="CONFIG_SMP"
+
+    if ! check_kernel_config "$CORE_CONFIGS"; then
+        log_fail "Core Scheduler kernel config validation failed"
+        pass=false
+    else
+        log_pass "Core Scheduler configs available"
+    fi
+
+    OPTIONAL_CONFIGS="CONFIG_SCHED_DEBUG CONFIG_CFS_BANDWIDTH CONFIG_NO_HZ_COMMON CONFIG_SCHED_AUTOGROUP CONFIG_CGROUP_SCHED CONFIG_CPU_FREQ_GOV_SCHEDUTIL CONFIG_FAIR_GROUP_SCHED"
+
+    if ! check_optional_config "$OPTIONAL_CONFIGS"; then
+        log_info "Optional Scheduler kernel configs are not available"
+    else
+        log_pass "Optional Scheduler configs available"
+    fi
+
+fi
+
+if [ "$pass" = "true" ]; then
+    log_pass "$TESTNAME : Test Passed"
+    echo "$TESTNAME PASS" > "$res_file"
+else
+    log_fail "$TESTNAME : Test Failed"
+    echo "$TESTNAME FAIL" > "$res_file"
+fi
+
+log_info "-------------------Completed $TESTNAME Testcase----------------------------"
+exit 0
--- a/Runner/suites/Kernel/Scheduler/PELT_load_tracking/run.sh
+++ b/Runner/suites/Kernel/Scheduler/PELT_load_tracking/run.sh
@@ -1,0 +1,181 @@
+#!/bin/sh
+
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Robustly find and source init_env
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+INIT_ENV=""
+SEARCH="$SCRIPT_DIR"
+while [ "$SEARCH" != "/" ]; do
+    if [ -f "$SEARCH/init_env" ]; then
+        INIT_ENV="$SEARCH/init_env"
+        break
+    fi
+    SEARCH=$(dirname "$SEARCH")
+done
+
+if [ -z "$INIT_ENV" ]; then
+    echo "[ERROR] Could not find init_env (starting at $SCRIPT_DIR)" >&2
+    exit 1
+fi
+
+if [ -z "${__INIT_ENV_LOADED:-}" ]; then
+    # shellcheck disable=SC1090
+    . "$INIT_ENV"
+fi
+
+# shellcheck disable=SC1090,SC1091
+. "$TOOLS/functestlib.sh"
+
+TESTNAME="PELT_load_tracking"
+test_path=$(find_test_case_by_name "$TESTNAME")
+cd "$test_path" || exit 1
+res_file="./$TESTNAME.res"
+
+# Kill any background load task on exit
+LOAD_PID=""
+# shellcheck disable=SC2317
+cleanup() {
+    if [ -n "$LOAD_PID" ]; then
+        kill "$LOAD_PID" 2>/dev/null || true
+        LOAD_PID=""
+    fi
+}
+trap cleanup EXIT INT TERM
+
+log_info "================================================================================"
+log_info "============ Starting $TESTNAME Testcase ======================================="
+log_info "================================================================================"
+log_info "Functional validation: PELT tracks CPU utilization under real scheduler load"
+
+if ! check_dependencies grep awk cat; then
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+
+pass="true"
+
+# ---------------------------------------------------------------------------
+# Prerequisite: /proc/schedstat must be present
+# ---------------------------------------------------------------------------
+if [ ! -f /proc/schedstat ]; then
+    log_fail "/proc/schedstat not found  -  cannot perform PELT load tracking test"
+    log_warn "Enable CONFIG_SCHEDSTATS in the kernel"
+    echo "$TESTNAME FAIL" > "$res_file"
+    exit 0
+fi
+
+# Helper: sum rq_cpu_time (field 8) across all CPU lines  -  total ns on-CPU
+get_total_rq_cpu_time() {
+    awk '/^cpu[0-9]/ {sum += $8} END {print sum+0}' /proc/schedstat
+}
+
+# Helper: sum run_delay (field 9)  -  total ns waiting in runqueue
+get_total_run_delay() {
+    awk '/^cpu[0-9]/ {sum += $9} END {print sum+0}' /proc/schedstat
+}
+
+# Helper: sum pcount (field 10)  -  total scheduling events
+get_total_pcount() {
+    awk '/^cpu[0-9]/ {sum += $10} END {print sum+0}' /proc/schedstat
+}
+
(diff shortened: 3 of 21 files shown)


smuppand:
@vnarapar please fix the workflow issues

smuppand (changes requested):
Since there are multiple files in the PR, I am providing a consolidated review.
 
1. CI is failing for executable permissions and Shell Lint.
2. All [run.sh](http://run.sh/) files use `[ -z "$__INIT_ENV_LOADED" ]`; please switch to `${__INIT_ENV_LOADED:-}`.
3. All [run.sh](http://run.sh/) files call check_dependencies but do not check the return value.
4. PELT_decay Method 1 reads `/proc/self/sched` through grep, so it reads the grep process sched data instead of the shell workload. Please read `/proc/$$/sched` instead.
5. PELT_sched_debug checks only whether `/sys/kernel/debug` exists, not whether debugfs is mounted. Please validate debugfs via `/proc/mounts`.
6. PELT_schedutil uses `-ge` for “frequency increased”, making the equal-frequency warning path unreachable. Use `-gt` for increase and handle equality separately.
7. Several scripts write .res and then exit 1. Please exit 0 after writing test FAIL/SKIP results so LAVA can publish the result cleanly.
8. PELT_config still has copy-paste “DMA-BUF” log messages.
9. Some tunables/sysfs entries are kernel-version/config dependent and should not be hard failures unless explicitly documented.
10. YAMLs currently mask cd/run/send-to-lava failures with `|| true`. Please remove unnecessary masking once [run.sh](http://run.sh/) result handling is fixed.
 
After these fixes and runtime evidence on at least one target, this can be reviewed again.

vnarapar:
Hi @smuppand , thanks fixed the issues. Have a query wrt Iusse#2. I see that most scripts are using **[ -z "$__INIT_ENV_LOADED" ];,** any reason to move to **${__INIT_ENV_LOADED:-}.**
Will share the runtime status once done.

smuppand:
> Hi @smuppand , thanks fixed the issues. Have a query wrt Iusse#2. I see that most scripts are using **[ -z "$__INIT_ENV_LOADED" ];,** any reason to move to **${__INIT_ENV_LOADED:-}.** Will share the runtime status once done.

because ${var:-} is safe when the variable is unset.

In the current script, the code directly expands "$__INIT_ENV_LOADED" before sourcing init_env. 

Why it matters

In normal /bin/sh without set -u, both may work.

But if the script is run by a stricter wrapper, CI harness, or future common runner with:

set -u

then this line can fail:

"$__INIT_ENV_LOADED"

because the variable may not exist yet.

You can get an error like:

__INIT_ENV_LOADED: parameter not set

Using this form avoids that:

"${__INIT_ENV_LOADED:-}"

It means:

Use $__INIT_ENV_LOADED if it is set; otherwise use empty string.

So the test becomes safe even when the variable is unset.

Why we use it in repo scripts

This pattern is already safer and more portable:

if [ -z "${__INIT_ENV_LOADED:-}" ]; then
    . "$INIT_ENV"
fi

It avoids:

Runner/suites/Kernel/Scheduler/PELT_config/run.sh
```diff
@@ -0,0 +1,84 @@
...
+test_path=$(find_test_case_by_name "$TESTNAME")
+cd "$test_path" || exit 1
+res_file="./$TESTNAME.res"
+
+log_info "================================================================================"
+log_info "============ Starting $TESTNAME Testcase ======================================="
+log_info "================================================================================"
+log_info "Validates kernel configuration required for PELT (Per-Entity Load Tracking)"
+
+if ! check_dependencies grep; then
```
smuppand: uses zgrep but dependency check only checks grep. Also better to log_skip with proper message to console for better debugging.

Runner/suites/Kernel/Scheduler/PELT_config/run.sh
```diff
@@ -0,0 +1,84 @@
...
+        pass=false
+    else
+        log_pass "Core Scheduler configs available"
+    fi
+
+    OPTIONAL_CONFIGS="CONFIG_SCHED_DEBUG CONFIG_CFS_BANDWIDTH CONFIG_NO_HZ_COMMON CONFIG_SCHED_AUTOGROUP CONFIG_CGROUP_SCHED CONFIG_CPU_FREQ_GOV_SCHEDUTIL"
+
+    log_info "Checking optional Scheduler configurations..."
+    for cfg in $OPTIONAL_CONFIGS; do
+        if zgrep -qE "^${cfg}=(y|m)" /proc/config.gz 2>/dev/null; then
```
smuppand: Please use check_optional_config function from functestlib.sh

Runner/suites/Kernel/Scheduler/PELT_config/run.sh
```diff
@@ -0,0 +1,84 @@
...
+
+pass="true"
+
+log_info "=== Core PELT / CFS Kernel Configs ==="
+
+if [ ! -f /proc/config.gz ]; then
+    log_warn "/proc/config.gz not found  -  kernel config checks will be skipped"
+    log_warn "Ensure CONFIG_IKCONFIG and CONFIG_IKCONFIG_PROC are enabled in the kernel"
+else
+    CORE_CONFIGS="CONFIG_FAIR_GROUP_SCHED CONFIG_SMP"
```
smuppand: Keep CONFIG_SMP required and move CONFIG_FAIR_GROUP_SCHED to optional unless the test is specifically validating group-scheduler PELT

Runner/suites/Kernel/Scheduler/PELT_sched_debug/run.sh
```diff
@@ -0,0 +1,194 @@
...
+pass="true"
+
+SCHED_DEBUG_DIR="/sys/kernel/debug/sched"
+
+# ---------------------------------------------------------------------------
+# Ensure debugfs is mounted; attempt to mount if root and not present
+# ---------------------------------------------------------------------------
+log_info "=== debugfs Mount Check ==="
+
+if ! grep -q "debugfs" /proc/mounts 2>/dev/null; then
```
smuppand: This only checks whether any debugfs mount exists, not whether debugfs is mounted at /sys/kernel/debug.
If debugfs is mounted somewhere else, the script may continue and then fail on /sys/kernel/debug/sched.

Runner/suites/Kernel/Scheduler/PELT_sched_debug/run.sh
```diff
@@ -0,0 +1,194 @@
...
+test_path=$(find_test_case_by_name "$TESTNAME")
+cd "$test_path" || exit 1
+res_file="./$TESTNAME.res"
+
+log_info "================================================================================"
+log_info "============ Starting $TESTNAME Testcase ======================================="
+log_info "================================================================================"
+log_info "Validates the scheduler debugfs interface used to inspect PELT state"
+
+if ! check_dependencies grep awk cat; then
```
smuppand: include id, mount, basename

Runner/suites/Kernel/Scheduler/PELT_tunables/run.sh
```diff
@@ -0,0 +1,185 @@
...
+        fi
+    else
+        log_fail "  $tunable not found at $path"
+        pass="false"
+    fi
+done
+
+# Validate ordering: latency >= min_granularity (CFS invariant)
+lat_path="$SYSCTL_DIR/sched_latency_ns"
+gran_path="$SYSCTL_DIR/sched_min_granularity_ns"
```
smuppand, line 73: Current script still fails if these are missing:
```
sched_min_granularity_ns
sched_latency_ns
sched_wakeup_granularity_ns
```
[run.sh](http://run.sh/)
These sysctls can be kernel-version/config dependent. The test can fail a valid target just because those tunables are not exposed.
Recommended fix: Either document these as mandatory for this test, or make missing tunables warning-only:

```log_warn "  $tunable not found at $path (kernel/config dependent)"```

Do not set pass="false" for absence unless this test is explicitly enforcing those knobs

**Review by smuppand — CHANGES_REQUESTED** [review]

**Comment by smuppand** [comment]
@vnarapar, please rebase and apply the small fixes mentioned above.

File: Runner/suites/Kernel/Scheduler/PELT_config/run.sh (hunk 3 above)
Comment (vnarapar) [reply Runner/suites/Kernel/Scheduler/PELT_config/run.sh]: Updated

File: Runner/suites/Kernel/Scheduler/PELT_config/run.sh (hunk 2 above)
Comment (vnarapar) [reply Runner/suites/Kernel/Scheduler/PELT_config/run.sh]: Updated

File: Runner/suites/Kernel/Scheduler/PELT_config/run.sh (hunk 1 above)
Comment (vnarapar) [reply Runner/suites/Kernel/Scheduler/PELT_config/run.sh]: Updated

File: Runner/suites/Kernel/Scheduler/PELT_sched_debug/run.sh (hunk 5 above)
Comment (vnarapar) [reply Runner/suites/Kernel/Scheduler/PELT_sched_debug/run.sh]: Updated

File: Runner/suites/Kernel/Scheduler/PELT_tunables/run.sh (hunk 6 above)
Comment (vnarapar) [reply Runner/suites/Kernel/Scheduler/PELT_tunables/run.sh:73]: Updated tunables to warn

File: Runner/suites/Kernel/Scheduler/PELT_sched_debug/run.sh (hunk 4 above)
Comment (vnarapar) [reply Runner/suites/Kernel/Scheduler/PELT_sched_debug/run.sh]: Updated

File: Runner/suites/Kernel/Scheduler/PELT_config/run.sh [hunk 7]
```diff
@@ -0,0 +1,83 @@
...
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+
+pass="true"
+
+log_info "=== Core PELT / CFS Kernel Configs ==="
+
+if [ ! -f /proc/config.gz ]; then
+    log_warn "/proc/config.gz not found  -  kernel config checks will be skipped"
```
smuppand, line 52: If /proc/config.gz is missing, the script only logs warnings and continues. Since pass="true" remains unchanged, the test later reports PASS. A target without /proc/config.gz can falsely pass PELT_config. The README says this condition should be SKIP.

Runner/suites/Kernel/Scheduler/PELT_sched_debug/run.sh
```diff
@@ -0,0 +1,194 @@
...
+test_path=$(find_test_case_by_name "$TESTNAME")
+cd "$test_path" || exit 1
+res_file="./$TESTNAME.res"
+
+log_info "================================================================================"
+log_info "============ Starting $TESTNAME Testcase ======================================="
+log_info "================================================================================"
+log_info "Validates the scheduler debugfs interface used to inspect PELT state"
+
+if ! check_dependencies grep awk cat id, mount, basename; then
```
smuppand: this test can incorrectly SKIP even when id and mount are installed. Please remove `,` separation.

Runner/suites/Kernel/Scheduler/PELT_sched_debug/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+```sh
+./run.sh
+```
+
+## Dependencies
+
+- `CONFIG_SCHED_DEBUG` kernel config
+- debugfs mounted at `/sys/kernel/debug`
+- `grep`, `awk`, `cat`
```
smuppand: The documentation does not match runtime requirements

Runner/suites/Kernel/Scheduler/PELT_config/run.sh
```diff
@@ -0,0 +1,83 @@
...
+    else
+        log_pass "Core Scheduler configs available"
+    fi
+
+    OPTIONAL_CONFIGS="CONFIG_SCHED_DEBUG CONFIG_CFS_BANDWIDTH CONFIG_NO_HZ_COMMON CONFIG_SCHED_AUTOGROUP CONFIG_CGROUP_SCHED CONFIG_CPU_FREQ_GOV_SCHEDUTIL CONFIG_FAIR_GROUP_SCHED"
+
+    if ! check_optional_config "$OPTIONAL_CONFIGS"; then
+        log_info "Optional Scheduler kernel config validation failed"
+    else
+        log_pass "Optional Scheduler configs available"
```
smuppand, line 71: These configs are optional. Logging "validation failed" is misleading when missing optional configs are warning-only. So remove the else block.

Runner/suites/Kernel/Scheduler/PELT_config/PELT_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - rb3gen2
+    - qcs6490
+    - qcs8300
+    - qcs9100
+    - sa8775p
+
+run:
+  steps:
+    - REPO_PATH=$PWD
+    - cd "$REPO_PATH/Runner/suites/Kernel/Scheduler/PELT_config"  || true
```
smuppand: cd ... || true masks path errors.

smuppand (changes requested):

Runner/suites/Kernel/Scheduler/PELT_sched_debug/run.sh
vnarapar: Updated

Runner/suites/Kernel/Scheduler/PELT_sched_debug/README.md
vnarapar: Updated

Runner/suites/Kernel/Scheduler/PELT_config/PELT_config.yaml
vnarapar: Updated

Runner/suites/Kernel/Scheduler/PELT_config/run.sh
vnarapar: Updated as below
log_info "Optional Scheduler kernel configs are not available"

Runner/suites/Kernel/Scheduler/PELT_config/run.sh
vnarapar: Updated

Runner/suites/Kernel/Scheduler/PELT_load_tracking/run.sh
```diff
@@ -0,0 +1,181 @@
...
+    echo "$TESTNAME SKIP" > "$res_file"
+    exit 0
+fi
+
+pass="true"
+
+# ---------------------------------------------------------------------------
+# Prerequisite: /proc/schedstat must be present
+# ---------------------------------------------------------------------------
+if [ ! -f /proc/schedstat ]; then
```
smuppand, line 62: /proc/schedstat is not PELT evidence

Rename and scope this as a schedstat activity test, or validate a PELT-specific signal such as se.avg.util_avg for the workload task when the required debug capability is available.

Runner/suites/Kernel/Scheduler/PELT_schedstat/run.sh
```diff
@@ -0,0 +1,174 @@
...
+
+total_runtime=$(awk '/^cpu[0-9]/ {sum += $8} END {print sum+0}' /proc/schedstat)
+total_delay=$(awk '/^cpu[0-9]/ {sum += $9} END {print sum+0}' /proc/schedstat)
+total_pcount=$(awk '/^cpu[0-9]/ {sum += $10} END {print sum+0}' /proc/schedstat)
+
+log_info "  Total rq_cpu_time : ${total_runtime} ns"
+log_info "  Total run_delay   : ${total_delay} ns"
+log_info "  Total pcount      : ${total_pcount}"
+
+if [ "$total_runtime" -gt 0 ] 2>/dev/null; then
```
smuppand, line 118: false PELT pass criterion. Keep this as an informational schedstat check, or move it to a scheduler-statistics suite. Do not make it the functional pass criterion for PELT.

Runner/suites/Kernel/Scheduler/PELT_sched_debug/run.sh
```diff
@@ -0,0 +1,194 @@
...
+
+# ---------------------------------------------------------------------------
+# Ensure debugfs is mounted; attempt to mount if root and not present
+# ---------------------------------------------------------------------------
+log_info "=== debugfs Mount Check ==="
+
+if ! grep -qE "debugfs /sys/kernel/debug" /proc/mounts 2>/dev/null; then
+    if [ "$(id -u 2>/dev/null)" = "0" ]; then
+        log_info "debugfs not mounted  -  attempting to mount..."
+        if mount -t debugfs debugfs /sys/kernel/debug 2>/dev/null; then
```
smuppand, line 59: test-mounted debugfs is never restored. Record whether the test mounted debugfs and unmount it in an EXIT, INT, and TERM cleanup handler only when the test mounted it.

Runner/suites/Kernel/Scheduler/PELT_schedstat/PELT_schedstat.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - rb3gen2
+    - qcs6490
+    - qcs8300
+    - qcs9100
+    - sa8775p
+
+run:
+  steps:
+    - REPO_PATH=$PWD
+    - cd "$REPO_PATH/Runner/suites/Kernel/Scheduler/PELT_schedstat"  || true
```
smuppand, line 21: masked directory failure. cd ... || true suppresses failure to enter the suite directory.

Remove || true and use the standard separate repo-path, cd, invocation, and result-upload steps.

PR #449, opened 2026-05-08
lava: improve testcase signal emission on serial console

smuppand:
This PR updates `Runner/utils/[send-to-lava.sh](http://send-to-lava.sh/)` to emit buffered LAVA testcase signals using shell builtin `printf` instead of `cat`.
 
The existing helper buffers LAVA signals into a temporary file and then emits them with:
 
```sh
cat "$SIGNAL_FILE"This PR updates `Runner/utils/[send-to-lava.sh](http://send-to-lava.sh/)` to emit buffered LAVA testcase signals using shell builtin `printf` instead of `cat`.

### DIFF
--- a/Runner/utils/send-to-lava.sh
+++ b/Runner/utils/send-to-lava.sh
@@ -28,10 +28,16 @@ else
     echo "[WARNING] Result file missing: $RESULT_FILE" >&2
 fi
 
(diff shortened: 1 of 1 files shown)


bhargav0610 (approved):
LGTM

PR #452, opened 2026-05-10
system: harden EFI variable validation for platform differences

smuppand:
This PR improves `EFI_Variable_Validation` to handle platform-specific EFI runtime behavior more robustly and introduces a reusable System-suite helper library.
 
The change adds:
 
- `Runner/utils/[lib_system.sh](http://lib_system.sh/)`
- Updated `Runner/suites/System/EFI_Variable_Validation/[run.sh](http://run.sh/)`
 
## Why this PR is needed
 
EFI variable behavior is firmware and platform dependent. In LAVA runs, the current EFI validation test can fail even when the kernel EFI configuration and EFI runtime filesystem are present.
 
Observed failures include:
 
```text
efivar: error listing variables: Function not implemented

### DIFF
--- a/Runner/suites/System/EFI_Variable_Validation/run.sh
+++ b/Runner/suites/System/EFI_Variable_Validation/run.sh
@@ -31,14 +31,19 @@ fi
 # shellcheck disable=SC1091
 . "$TOOLS/functestlib.sh"
 
+# shellcheck disable=SC1091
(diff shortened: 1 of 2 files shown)


bhargav0610 (commented):
LGTM

bhargav0610 (approved):

PR #462, opened 2026-05-28
irq: make arch_timer validation robust under tickless idle

smuppand:
This PR makes the `irq` baseport test robust against false LAVA failures caused by passive `arch_timer` counter sampling.
 
The previous test logic sampled the `arch_timer` interrupt line, slept for a fixed window, and failed if every CPU counter did not increment. That assumption is weak on modern Linux systems because an online CPU may remain idle or tickless during the sampling window due to:
 
- NO_HZ / tickless idle
- CPU idle states
- scheduler not placing work on a CPU
- CPU isolation
- affinity or cpuset restrictions
- low system activity during LAVA runs
- platform power management
 
This could produce false failures such as:
 
```text
CPU3 initial=3764 final=3764 delta=0
```

lava job for reference with these changes. https://lava.infra.foundries.io/scheduler/job/236802#L2423

### DIFF
--- a/Runner/suites/Kernel/Baseport/irq/irq.yaml
+++ b/Runner/suites/Kernel/Baseport/irq/irq.yaml
@@ -1,7 +1,7 @@
 metadata:
     format: Lava-Test Test Definition 1.0
-    name: AUSanity
-    description: "Baseline AU Sanity test plan"
+    name: irq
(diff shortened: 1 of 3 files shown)


abbajaj806 (approved):
LGTM

PR #463, opened 2026-05-29
hotplug: validate CPU hotplug using runtime-discovered topology

smuppand:
Rework the CPU hotplug validation to scale across Qualcomm platforms with different CPU counts, capacities, and cluster layouts.
 
The previous test used a fixed CPU range and attempted to offline CPUs directly. This was not scalable across SoCs and could produce weak or misleading failures when CPU topology differed from the assumed layout.
 
This PR updates the test to discover CPU topology dynamically at runtime and validate only CPUs that are actually exposed as hotplug-controllable through sysfs.

Please refer this lava job https://lava.infra.foundries.io/scheduler/job/236862 for the results with these patches.


abbajaj806 (approved):
LGTM

PR #464, opened 2026-05-29
ci: add weekly qcom-linux-testkit snapshot tag workflow

smuppand:
Add a GitHub Actions workflow to create annotated qcom-linux-testkit source snapshot tags from `main` using the agreed CalVer tag convention:
 
```text
[testkit-YYYY.MM](http://testkit-yyyy.mm/).DD


abbajaj806 (approved):
LGTM

PR #467, opened 2026-05-31
shmbridge: avoid false failures from qcom_scm kernel cmdline

smuppand:
The shmbridge test scans current-boot kernel logs for qcom_scm-related errors. The existing match is too broad and can treat benign boot arguments such as qcom_scm.download_mode=1 as failures when other kernel command line parameters contain error-like words such as panic.

Update the log scan to:
- collect qcom_scm/shmbridge-related lines
- ignore Kernel command line entries
- ignore qcom_scm.download_mode
- ignore CONFIG_QCOM_SCM references
- fail only when the remaining lines contain real error keywords

This avoids false LAVA failures while preserving detection of genuine qcom_scm or shmbridge errors.


abbajaj806 (approved):
LGTM

PR #481, opened 2026-06-04
OpenCV: fix LAVA YAML flag argument handling

smuppand:
Fix `OpenCV.yaml` argument handling so the OpenCV LAVA job no longer passes values to flag-style `[run.sh](http://run.sh/)` options.
 

The current YAML passes boolean values directly to flag options:
 
```sh
--list "${LIST_ONLY}"
--shuffle "${SHUFFLE}"


abbajaj806 (approved):
LGTM

PR #482, opened 2026-06-04
WiFi_OnOff: improve wait handling for delayed interface creation

smuppand:
Improve WiFi interface discovery robustness for `WiFi_OnOff` by allowing more time for delayed WiFi netdev creation in LAVA/CI environments.
 
Why
 
`WiFi_OnOff` can fail in LAVA even when the WiFi stack is present:
 
- required WiFi kernel configs are enabled
- WiFi-related modules are loaded
- no WiFi probe/runtime failures are detected
- but no `wlan*`/WiFi interface appears within the current wait window
 
This can happen because driver/module visibility, firmware initialization, cfg80211 PHY registration, and netdev creation are asynchronous. In automated CI, the interface may appear later than the current 30-second wait.


bhargav0610 (dismissed):

qcom-anilyada:
Failed job from the pre-merge workflow: https://lava.infra.foundries.io/scheduler/job/243672

Job Without PR: https://lava.infra.foundries.io/scheduler/job/244641
Job with PR: https://lava.infra.foundries.io/scheduler/job/244642

smuppand:
@qcom-anilyada and @bhargav0610 I'm noticing a pattern with the device connected to [rb4-hyd-01](https://lava.infra.foundries.io/scheduler/device/rb4-hyd-01).  Please check—every time, the wifi test fails because the interface isn't coming up at all.
https://lava.infra.foundries.io/results/244787
https://lava.infra.foundries.io/results/244780

The same test passes on other RB4 machines.
https://lava.infra.foundries.io/scheduler/job/244784
https://lava.infra.foundries.io/scheduler/job/244783
https://lava.infra.foundries.io/scheduler/job/244781
https://lava.infra.foundries.io/scheduler/job/244778

Runner/utils/lib_connectivity.sh
```diff
@@ -19,39 +19,68 @@ wait_for_wifi_interface() {
     sleep_step="${2:-2}"
     waited=0
     iface=""
-
+    settle_done=0
+ 
     case "$max_wait" in
         ''|*[!0-9]*)
             max_wait=30
```
azharsha: Should this be updated to 60 too?

Runner/utils/lib_connectivity.sh
```diff
@@ -19,39 +19,68 @@ wait_for_wifi_interface() {
...
             ;;
     esac
-
+
```
azharsha, line 29: Extra white-space

Runner/utils/lib_connectivity.sh
```diff
@@ -19,39 +19,68 @@ wait_for_wifi_interface() {
...
     esac
-
+ 
     case "$sleep_step" in
         ''|*[!0-9]*)
             sleep_step=2
             ;;
     esac
-
+
```
azharsha, line 35: Extra white-space

Runner/utils/lib_connectivity.sh
```diff
@@ -19,39 +19,68 @@ wait_for_wifi_interface() {
...
+                printf '%s\n' "$iface" >&3
+                return 0
+            fi
+ 
+            if [ "$settle_done" -eq 0 ]; then
+                settle_done=1
+ 
+                if command -v udevadm >/dev/null 2>&1; then
+                    log_info "No WiFi interface yet; triggering net uevents and waiting for udev settle"
+                    udevadm trigger --subsystem-match=net >/dev/null 2>&1 || true
```
azharsha: This will re-trigger on all network interfaces, which are already settled.
Perhaps this can be done

`udevadm trigger --action=add --subsystem-match=net >/dev/null 2>&1 || true`

Runner/utils/lib_connectivity.sh
```diff
@@ -19,39 +19,68 @@ wait_for_wifi_interface() {
...
-        fi
-
-        sleep "$sleep_step"
-        waited=$((waited + sleep_step))
-    done
-
+ 
+    # Keep stdout reserved for the detected interface name. This helper is
+    # commonly used through command substitution, so all diagnostics from the
+    # wait loop must go to stderr. The final interface is printed through fd 3,
```
azharsha: Should the README be also updated to let the user know that fd 3 is in use and they cannot use it.

Runner/utils/lib_connectivity.sh
smuppand: > Should this be updated to 60 too?

Addressed. I updated the helper default and fallback wait values to 60s so direct callers of wait_for_wifi_interface() behave consistently with WiFi_OnOff/[run.sh](http://run.sh/).

Runner/utils/lib_connectivity.sh
smuppand: > This will re-trigger on all network interfaces, which are already settled. Perhaps this can be done
> 
> `udevadm trigger --action=add --subsystem-match=net >/dev/null 2>&1 || true`

Addressed. Updated the trigger to use --action=add with --subsystem-match=net so it is more targeted.

Runner/utils/lib_connectivity.sh
smuppand: > Should the README be also updated to let the user know that fd 3 is in use and they cannot use it.

Good point. I removed the fd 3 usage completely, so no README update is needed. The helper now sends diagnostics explicitly to stderr and keeps stdout reserved only for the detected interface name.

bhargav0610 (approved):

PR #486, opened 2026-06-08
Detect whitespace

azharsha:


.github/workflows/detect-trailing-whitespace.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check for trailing whitespace
+        run: |
+          set -eu
+          git diff --check origin/main...HEAD \
```
smuppand: NACK

smuppand (changes requested):
- Move `git diff --check` into the existing Shell Lint/Preflight workflow.
- Keep it pull_request-focused.
- Split repository-wide whitespace cleanup into a separate mechanical PR only if maintainers want it.

azharsha:
> * Move `git diff --check` into the existing Shell Lint/Preflight workflow.
> * Keep it pull_request-focused.
ACK
> * Split repository-wide whitespace cleanup into a separate mechanical PR only if maintainers want it.
Why would maintainers not want to remove whitespace cleanup? As mentioned in the commit message, there is no functional change attached with the cleanup. Please list the maintainers I will add them to this PR.

smuppand:
Please rebase. Additionally, share the sandbox results if validated with your branch. Since the changes will be added to the workflow.

azharsha:
> Please rebase. Additionally, share the sandbox results if validated with your branch. Since the changes will be added to the workflow.

Rebased. Test workflow run. https://github.com/azharsha/qcom-linux-testkit/actions/runs/27803460389/job/82278371361

smuppand (changes requested):
1. What specific whitespace issues does this detect?
2. Show examples of bugs this would have caught
3. Why not use .editorconfig + pre-commit hooks?
4. Can this be integrated into qcom-preflight-checks?

Justify why this needs to be a workflow instead of using standard tools like `.editorconfig` and `pre-commit` hooks.

azharsha:
> 1. What specific whitespace issues does this detect?
> 2. Show examples of bugs this would have caught

That should be visible in the commit. Do you want me to add examples to the commit message? Is that the ask? Please clarify.


> 3. Why not use .editorconfig + pre-commit hooks?

Answered below.



> 4. Can this be integrated into qcom-preflight-checks?

Would prefer to keep it in shellcheck, looking at qcom-preflight-checks does not look the correct place to keep it.
It is doing very basic checks related to the PR. 

> Justify why this needs to be a workflow instead of using standard tools like `.editorconfig` and `pre-commit` hooks.

Having this as a workflow enforces the check. Developers can bypass local checks using --no-verify.

PR #487, opened 2026-06-10
OpenCV: add bounded smoke binary list for CI

smuppand:
Limit the default OpenCV CI execution to a bounded smoke set instead of running the full discovered OpenCV suite.
 
The OpenCV YAML currently defaults to:
 
- `BIN_PATH=""`
- `SUITE=all`
 
This causes `[run.sh](http://run.sh/)` to discover and run all `opencv_test_*` and `opencv_perf_*` binaries. In LAVA jobs with a shared `lava-test-shell` timeout, the full OpenCV suite can consume the shell timeout before `OpenCV.res` is written. When that happens, LAVA marks OpenCV and all following test definitions as missing failures.
 
This change adds a repo-standard `--bin-list` mode in `OpenCV/[run.sh](http://run.sh/)` and keeps the YAML run steps simple.
 
Default CI smoke binaries:
 
- `opencv_test_core`
- `opencv_test_sfm`

Reference job with the current patch - https://lava.infra.foundries.io/scheduler/job/255904

### DIFF
--- a/Runner/suites/Multimedia/OpenCV/OpenCV.yaml
+++ b/Runner/suites/Multimedia/OpenCV/OpenCV.yaml
@@ -9,22 +9,23 @@ metadata:
 
 params:
   BIN_PATH: "" # --bin <path|name>
+  BIN_LIST: "opencv_test_sfm" # --bin-list "bin1 bin2 ..."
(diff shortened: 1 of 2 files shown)


abbajaj806 (approved):
LGTM

PR #490, opened 2026-06-12
WiFi_OnOff: add bounded recovery for missing WiFi interface

smuppand:
Improve WiFi_OnOff robustness for intermittent cases where WiFi modules are loaded but no WiFi interface is exposed before the wait timeout.

The test can intermittently fail in LAVA even though WiFi kernel configs and modules are present. In these cases, no usable WiFi runtime object is exposed:
no WiFi interface from get_wifi_interface
empty iw dev
empty iw phy
empty /sys/class/ieee80211

Sample lava job for reference https://lava.infra.foundries.io/scheduler/job/263091

### DIFF
--- a/Runner/utils/lib_connectivity.sh
+++ b/Runner/utils/lib_connectivity.sh
@@ -14,58 +14,88 @@ wifi_unblock_rfkill() {
 
 # Retry WiFi interface discovery for a bounded time while unblocking rfkill.
 # Prints interface name on success and returns non-zero on timeout.
+#
+# Optional recovery knobs:
+# WIFI_RECOVERY_RELOAD=1|0
+# Enable one controlled WiFi driver reload if no interface appears.
+#
+# WIFI_RECOVERY_RELOAD_AFTER_S=N
+# Try reload after N seconds of waiting. Default: half of max_wait.
+#
+# WIFI_RECOVERY_MODULES="mod1 mod2 ..."
+# Leaf WiFi driver modules eligible for reload.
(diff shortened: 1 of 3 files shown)


Runner/utils/lib_connectivity.sh
```diff
@@ -14,74 +14,149 @@ wifi_unblock_rfkill() {
...
+            for mod in $recovery_modules; do
+                [ -n "$mod" ] || continue
+
+                if ! is_module_loaded "$mod"; then
+                    continue
+                fi
+
+                log_warn "No WiFi interface after ${waited}s; attempting one controlled reload of $mod" >&2
+
+                if modprobe -r "$mod" >/dev/null 2>&1; then
```
vnarapar, line 122: Why do test script reload the module here? Isnt it needs to be taken care by Kernel to safely loada modules on bootup? or by udev?

smuppand: > Why do test script reload the module here? Isnt it needs to be taken care by Kernel to safely loada modules on bootup? or by udev?

we do not reload modules during the normal path.

1. Wait for WiFi interface normally
   - rfkill unblock
   - get_wifi_interface retry
   - udev trigger/settle

2. If interface appears
   - no module reload
   - continue toggle test

3. Only if interface does not appear after part of the wait
   - attempt one controlled reload
   - only for an already-loaded WiFi leaf driver module

bhargav0610 (approved):

PR #506, opened 2026-07-03
fix: Update Display test results to .res file

vnarapar:
Fix Display testscripts to update the PASS/FAIL status to result file

### DIFF
--- a/Runner/suites/Multimedia/Display/weston-clickdot/run.sh
+++ b/Runner/suites/Multimedia/Display/weston-clickdot/run.sh
@@ -280,8 +280,10 @@ echo "$TESTNAME $final" >"$RES_FILE"
 
 if [ "$final" = "PASS" ]; then
   log_pass "${TESTNAME} : PASS"
+  echo "${TESTNAME} PASS" >"$RES_FILE"
   exit 0
 fi
 
 log_fail "${TESTNAME} : FAIL"
+echo "${TESTNAME} FAIL" >"$RES_FILE"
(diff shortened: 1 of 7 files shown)


Runner/suites/Multimedia/Display/weston-clickdot/run.sh
```diff
@@ -280,8 +280,10 @@ echo "$TESTNAME $final" >"$RES_FILE"
 
 if [ "$final" = "PASS" ]; then
   log_pass "${TESTNAME} : PASS"
+  echo "${TESTNAME} PASS" >"$RES_FILE"
```
smuppand, line 283: This PASS write duplicates the existing final result write at line 279, and the FAIL path duplicates it again at line 288.

smuppand (changes requested):
This PASS write duplicates the existing final result write at line 279, and the FAIL path duplicates it again at line 288 across the files it is same. So i would suggest it is a duplicate result updation.

Runner/suites/Multimedia/Display/weston-clickdot/run.sh
smuppand: > This PASS write duplicates the existing final result write at line 279, and the FAIL path duplicates it again at line 288.

@vnarapar Can you relook?

vnarapar: Sry my bad. i will check and update

PR #518, opened 2026-07-24
video: enhance GStreamer encode/decode test framework

nitinn22:
## Summary
Enhance the GStreamer video encode/decode test framework.

## Changes

- Extend lib_gstreamer helper functionality
- Add reusable video helper support
- Improve framework integration
- Enhance encode/decode test execution flow and additional test cases added.
- Improve validation and artifact handling
- Update README documentation

## Validation
- Rebased on latest upstream/main
- Merge conflicts resolved with latest framework
- Encode/decode workflow validated
- Branch contains only video-related changes


Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -79,6 +82,7 @@ total_tests=0
 
 # -------------------- Defaults (LAVA env vars -> defaults; CLI overrides) --------------------
 testMode="${VIDEO_TEST_MODE:-all}"
+testType="${VIDEO_TEST_TYPE:-all}"
```
smuppand, line 85: This is a backward-incompatible change in runtime, hardware prerequisites, display requirements, network/clip requirements, and resource consumption. An existing nightly or LAVA job can change behavior without modifying its job definition.

Preserve compatibility by defaulting to "basic":

`testType="${VIDEO_TEST_TYPE:-basic}"`

Then add "VIDEO_TEST_TYPE" and "VIDEO_CONCURRENCY_SESSIONS" to "Video_Encode_Decode.yaml" and pass them explicitly. Dedicated advanced YAMLs may be added when separate LAVA testcase visibility is required.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,63 +690,993 @@ run_decode_test() {
...
+    log_info "Skipping other tests (test type: advanced-encode)"
+    ;;
+  all)
+    # Run ALL tests (new default behavior)
+    log_info "Running ALL video tests (basic + advanced)"
+    
+    # Run basic tests (will be executed below in normal flow)
+    log_info "Will run: basic encode/decode tests"
+    
+    # Run UVC test
```
smuppand, line 1527: The default "all" run does not actually exercise DRC or concurrency on its first execution. The downloaded files only become useful on a subsequent run, creating state-dependent results.

Reorder the main flow:

parse and validate
ensure stack
resolve selected groups
prepare all required clips
preflight hardware/plugins/display
execute selected groups
scan dmesg
emit summary

No clip-dependent testcase should run before "get_test_clip()" has completed for that group.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,63 +690,993 @@ run_decode_test() {
...
+    ;;
+  *)
+    log_warn "Unknown test type: $testType"
+    echo "$RESULT_TESTNAME SKIP" >"$RES_FILE"
+    exit 0
+    ;;
+esac
+
+# Skip basic tests if we ran a specific test type (not 'basic' or 'all')
+if [ "$testType" != "basic" ] && [ "$testType" != "all" ]; then
```
smuppand, line 1556: Put basic clip preparation and both basic loops inside an actual guard

case "$testType" in
    basic|all)
        prepare_basic_clips
        run_basic_tests
        ;;
esac

Alternatively, route each selected test type through one dispatcher and proceed directly to common dmesg/summary handling after it completes.

Also define how "--mode" interacts with advanced groups. Currently "--mode decode --test-type all" still executes all advanced encode tests.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,63 +690,993 @@ run_decode_test() {
...
+    log_skip "$testname: Failed to set up Wayland environment"
+    skip_count=$((skip_count + 1))
+    return 1
+  fi
+  
+  test_log="$OUTDIR/${testname}.log"
+  : >"$test_log"
+  
+  # Build pipeline using library function
+  pipeline=$(gstreamer_build_uvc_preview_pipeline "$uvc_dev" "1920" "1080" "5")
```
smuppand, line 829: UVC detection only establishes that a uvcvideo node exists, then unconditionally requests 1920x1080@5. Many valid UVC cameras do not expose that raw mode, so this becomes a false regression failure. Discover a supported mode dynamically, log the selected device/caps, and offer explicit device/width/height/framerate overrides with CLI-over-environment precedence.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,63 +690,993 @@ run_decode_test() {
...
+    log_fail "$testname: FAIL (rc=$gstRc)"
+    fail_count=$((fail_count + 1))
+    return 1
+  fi
+}
+
+# -------------------- H.264 Concurrency Decode Test --------------------
+# Test: H264_Decode_Concurrency_16x480p (PR43865)
+run_h264_concurrency_test() {
+  testname="H264_Decode_Concurrency_16x480p"
```
smuppand, line 899: Any ordinary decodable H.264 clip—or a no-error pipeline that times out—can satisfy the current PASS criteria. The testcase therefore does not prove the FR82787 dynamic-resolution-change requirement.

Capture verbose caps or decoder pad-cap events and require:

1. the expected initial resolution;
2. the expected second resolution;
3. the transition in the correct order;
4. frame/FPS evidence after the second caps event.

Write the observed transition to the testcase log and fail when either resolution or post-transition activity is missing.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,63 +690,993 @@ run_decode_test() {
...
-        else
-          log_warn "VP9 WebM clip not found in downloaded content"
-        fi
-      else
-        log_warn "VP9 WebM clip download failed (offline or URL issue)"
-      fi
+# -------------------- H.265 Concurrency Decode Test --------------------
+# Test: H265_Decode_Concurrency_16x1080p (PR43866)
+run_h265_concurrency_test() {
+  testname="H265_Decode_Concurrency_16x1080p"
```
smuppand, line 993: H265_Decode_Concurrency_16x1080p but the implementation uses H265_480p_30fps.mp4

The README describes this as a 480p test, while the CLI help describes it as 1080p.

Choose the actual requirement and make all four sources agree:

- testcase name;
- clip filename/content;
- help output;
- README.

For 1080p coverage, use a verified 1920×1080 clip and validate negotiated caps. Otherwise rename the testcase to "H265_Decode_Concurrency_16x480p".

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,63 +690,993 @@ run_decode_test() {
...
+  #   video_0::type=video ! video/x-raw,format=NV12,width=1280,height=720,framerate=30/1 ! queue ! scb.sink 
+  #   qtismartvencbin default-gop=30 max-gop=600 smart-framerate=true smart-gop=false max-bitrate=4200000 name=scb encoder=v4l2h265enc ! queue ! h265parse ! queue ! mp4mux ! queue ! filesink location=...
+  #   camsrc. ! video/x-raw,format=NV12,width=640,height=480,framerate=15/1 ! queue ! scb.sink_ctrl
+  pipeline="qtiqmmfsrc camera=0 noise-reduction=2 video_0::extra-buffers=20 video_0::type=preview name=camsrc video_0::type=video ! video/x-raw,format=NV12,width=1280,height=720,framerate=30/1 ! queue ! scb.sink qtismartvencbin default-gop=30 max-gop=600 smart-framerate=true smart-gop=false max-bitrate=4200000 name=scb encoder=v4l2h265enc ! queue ! h265parse ! queue ! mp4mux ! queue ! filesink location=${output_file} camsrc. ! video/x-raw,format=NV12,width=640,height=480,framerate=15/1 ! queue ! scb.sink_ctrl"
+  
+  log_info "Pipeline: $pipeline"
+  
+  # Run encoding
+  if gstreamer_run_gstlaunch_timeout "$((duration + 10))" "$pipeline" >>"$test_log" 2>&1; then
+    gstRc=0
```
smuppand, line 1234: Before every advanced encode:

rm -f "$output_file"

After execution, require:

- "gstRc" is an expected completion code, normally "0" or the controlled timeout value;
- the file was created by the current run;
- the file exceeds the minimum size;
- "gst-discoverer-1.0" or an equivalent helper confirms the expected container, codec, resolution, and duration.

Apply this to smart encode, cyclic IR, slice MB, and Rotate90.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1682,3 +1683,198 @@ camera_setup_wayland_environment() {
...
+  done
+  
+  printf '%s' "$pipeline"
+}
+
+# gstreamer_build_smart_encode_pipeline
+# Build pipeline for HEVC Smart Encode with dual camera streams
+# Args: output_file
+# Returns: pipeline string or empty on error
+gstreamer_build_smart_encode_pipeline() {
```
smuppand, line 1830: this function duplicates the complete smart-encode pipeline directly in the runner. The cyclic, slice, and rotate functions similarly bypass the new "gstreamer_build_camera_encode_pipeline()" helper.

Fixes to properties, stack handling, quoting, encoder controls, or muxing can be applied to one copy but not the other. The helper API is not the source of truth.

Build the pipelines through the shared helpers:

pipeline="$(gstreamer_build_smart_encode_pipeline "$output_file")"

and:

pipeline="$(
    gstreamer_build_camera_encode_pipeline \
        h265 3840 2160 "$output_file" \
        "controls,video_bitrate=8000000,intra_refresh_period_type=1" \
        "$detected_stack"
)"

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,63 +690,993 @@ run_decode_test() {
...
+  
+  log_fail "$testname: FAIL (no valid output)"
+  fail_count=$((fail_count + 1))
+  return 1
+}
+
+# -------------------- HEVC 4K Rotate90 Encode Test --------------------
+# Test: HEVC_Encode_4K_Rotate90 (FR72846)
+run_rotate_encode_test() {
+  testname="HEVC_Encode_4K_Rotate90"
```
smuppand, line 1412: "HEVC_Encode_4K_Rotate90" removes "rotate=90" when the detected stack is not downstream and then continues to validate the resulting ordinary encode as the Rotate90 testcase.

Treat rotation capability as a prerequisite:

if [ "$detected_stack" != "downstream" ]; then
    log_skip "$testname: rotate=90 control is unavailable on this stack"
    skip_count=$((skip_count + 1))
    return 1
fi

Also verify the rotation result, for example through expected output dimensions/caps or another reliable bitstream/output check. Do not silently downgrade the testcase to a normal 4K encode.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1682,3 +1683,198 @@ camera_setup_wayland_environment() {
...
+    printf 'v4l2src device=%s ! video/x-raw,width=%s,height=%s,framerate=%s/1 ! waylandsink fullscreen=true' \
+      "$device" "$width" "$height" "$framerate"
+  fi
+}
+
+# gstreamer_build_drc_decode_pipeline
+# Build pipeline for Dynamic Resolution Change H.264 decode test
+# Args: clip_path video_stack
+# Returns: pipeline string or empty on error
+gstreamer_build_drc_decode_pipeline() {
```
smuppand, line 1749: The new builders bypass the stack-aware behavior already used by "gstreamer_build_v4l2_decode_pipeline()".

Tests advertised for general execution can fail or use unsupported properties on upstream drivers.
Centralize stack-dependent decoder arguments in one helper, for example:

gstreamer_v4l2_decoder_args() {
    case "$1" in
        downstream) printf '%s\n' "capture-io-mode=4 output-io-mode=4" ;;
        *) printf '\n' ;;
    esac
}

Use it in the generic, DRC, and concurrency builders. Pass "$detected_stack" into the concurrency builder as well.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1682,3 +1683,198 @@ camera_setup_wayland_environment() {
...
+  clip_path="$2"
+  sessions="$3"
+  
+  [ -n "$codec" ] || return 1
+  [ -n "$clip_path" ] || return 1
+  [ -f "$clip_path" ] || return 1
+  [ -n "$sessions" ] || sessions="16"
+  
+  # Build qtivcomposer with 4x4 grid layout
+  pipeline="qtivcomposer name=mix"
```
smuppand, line 1775: Validate a supported range at both the CLI and helper boundary, such as "1..16", and dynamically generate only the required composer pads and positions. Generate the grid from the requested session count or explicitly support only "16" and remove the configurable option.

The result/test name should also include the actual count:

testname="H264_Decode_Concurrency_${concurrencySessions}x480p"

smuppand (changes requested):
Move the functions out of run.sh to lib_gstreamer.sh. I see duplicates and still optimizations can be done. 

The current commit carries ~1600 lines of code, split it into logical commits inside the same PR:

1. "gstreamer: add advanced video pipeline helpers"
   
   - UVC discovery;
   - stack-aware decoder arguments;
   - advanced pipeline builders;
   - common execution and validation helpers.

2. "video: add UVC and DRC GStreamer tests"
   
   - UVC preview;
   - DRC caps-transition validation;
   - display prerequisites.

3. "video: add concurrent decode tests"
   
   - H.264/H.265/MJPEG;
   - dynamic session layout;
   - concurrency-specific validation.

4. "video: add downstream advanced encode tests"
   
   - smart encode;
   - cyclic IR;
   - slice MB;
   - Rotate90;
   - current-run artifact validation.

5. "docs: document advanced GStreamer video coverage"
   
   - synchronized defaults;
   - prerequisites;
   - exact PASS/SKIP criteria;
   - verified execution examples.

Runner/utils/lib_gstreamer.sh
nitinn22: making this limited and set to 8 x 480p sessions to support on all targets.

nitinn22:
@smuppand , followed similar pattern like earlier all runner script related changes in one commit and helper function changes in another. addressed the review comments and some comments updated explicitly.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1367,33 +1382,149 @@ run_pipeline_with_logs() {
...
+                        ;;
+                    mjpeg)
+                        decoder_count=$(grep -c "jpegdec.*negotiated\|avidemux.*negotiated" "$console_log" || echo "0")
+                        ;;
+                    *)
+                        decoder_count=0
+                        ;;
+                esac
+                
+                if [ "$decoder_count" -ge "$expected_count" ]; then
```
smuppand: When fewer decoder branches than expected_count are observed, the code only logs a warning. Execution then reaches line 1495 and returns success 

Return failure when decoder_count < expected_count. Normalize the count with grep -c ... || true; do not use || echo 0, because grep -c already prints zero.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1367,33 +1382,149 @@ run_pipeline_with_logs() {
...
+                # Extract all video caps with resolution
+                caps_lines=$(grep -E "caps.*video/x-raw.*width.*height" "$console_log" || true)
+                
+                if [ -z "$caps_lines" ]; then
+                    log_fail "${name}: No video caps detected - cannot verify DRC"
+                    return 1
+                fi
+                
+                # Count distinct resolutions (look for 1920x1080 and 1280x720)
+                has_1080p=$(printf '%s\n' "$caps_lines" | grep -c "width.*1920.*height.*1080\|width=1920.*height=1080\|width=(int)1920.*height=(int)1080" || echo "0")
```
smuppand: grep -c ... || echo 0 produces two zero lines when there is no match. Numeric tests receive 0\n0, fail to parse, and the missing-resolution branches can be bypassed.

Use grep -c ... || true, sanitize each result as one integer, and require both resolutions before returning success.

Runner/utils/lib_gstreamer.sh
smuppand, line 2265: This path can start or restart Weston through weston_pick_env_or_start, but the runner cleanup does not track or restore that state

Track whether Weston was started or replaced and restore it through the existing lib_display.sh cleanup/restoration pattern on EXIT, INT and TERM.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,111 +668,986 @@ run_decode_test() {
...
+    log_skip "$testname: Failed to set up Wayland environment"
+    skip_count=$((skip_count + 1))
+    return 1
+  fi
+  
+  test_log="$OUTDIR/${testname}.log"
+  : >"$test_log"
+  
+  # Build pipeline using library function with 2 sessions
+  pipeline=$(gstreamer_build_concurrency_decode_pipeline "mjpeg" "$mjpeg_clip" "2")
```
smuppand: The helper’s third argument is video_stack, not session count, and the helper hardcodes sessions=8 

 Add an explicit session-count argument to gstreamer_build_concurrency_decode_pipeline, generate an appropriate two-session layout, and pass "$detected_stack" 2.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/README.md
```diff
@@ -75,6 +75,187 @@ By default, the test runs the following test cases at 4K resolution for H.264/H.
...
+- **Requirements**: qtivcomposer, Wayland display
+- **Test Clip**: `H265_480p_30fps.mp4`
+- **Pipeline (per session)**: `filesrc → qtdemux → h265parse → v4l2h265dec → qtivcomposer → waylandsink`
+
+**3. MJPEG_Decode_Concurrency_2x1080p (FR98277)**
+- **Description**: 2 concurrent MJPEG 1080p decode sessions
+- **Requirements**: jpegdec (software decoder), qtivcomposer, Wayland display
+- **Test Clip**: `mjpeg1.avi`
+- **Pipeline (per session)**: `filesrc → avidemux → jpegdec → qtivcomposer → waylandsink`
+
```
smuppand, line 129: Remove trailing whitespace from all changed files

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
nitinn22: session count want to limit to 8 sessions for h264, h265 and 2 sessions to mjpeg. made changes acc.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,111 +668,986 @@ run_decode_test() {
...
+  log_info "Exit code: $gstRc"
+  
+  # Validate log
+  if ! gstreamer_validate_log "$test_log" "$testname"; then
+    log_fail "$testname: FAIL (GStreamer errors detected)"
+    fail_count=$((fail_count + 1))
+    return 1
+  fi
+  
+  # Verify rotation: 90° rotation should swap dimensions (3840x2160 → 2160x3840)
```
smuppand: Remove the rotation block from this function. Keep rotation verification only in run_rotate_encode_test and validate cyclic-IR output as 3840x2160.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
smuppand: > session count want to limit to 8 sessions for h264, h265 and 2 sessions to mjpeg. made changes acc.

Add an explicit session-count argument to gstreamer_build_concurrency_decode_pipeline, validate it, generate the matching layout, and call MJPEG with "$detected_stack" "2".

Runner/utils/lib_gstreamer.sh
smuppand: Return failure when decoder_count < expected_count. Count distinct decoder instances or negotiated sink pads rather than relying only on *negotiated text.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1682,3 +1813,409 @@ camera_setup_wayland_environment() {
...
+  esac
+  
+  # Verify resolution
+  if grep -q "Width: $expected_width" "$discover_log" && grep -q "Height: $expected_height" "$discover_log"; then
+    log_info "$testname: ✓ Resolution: ${expected_width}x${expected_height}"
+  else
+    log_warn "$testname: Resolution verification: Expected ${expected_width}x${expected_height}"
+  fi
+  
+  # Verify duration (at least min_duration seconds)
```
smuppand: Parse duration into seconds, enforce min_duration, and return failure for required codec, resolution, or container mismatches.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/README.md
```diff
@@ -528,6 +766,89 @@ This test uses reusable helper functions from `lib_gstreamer.sh` that other GStr
...
+- Returns: Complete pipeline string with fpsdisplaysink wrapper
+- Example:
+  ```sh
+  pipeline=$(gstreamer_build_drc_decode_pipeline "/tmp/drc_clip.mp4" "downstream")
+  gstreamer_run_gstlaunch_timeout 60 "$pipeline"
+  ```
+
+#### 7. `gstreamer_build_concurrency_decode_pipeline`
+
+Build pipeline for concurrent decode tests with qtivcomposer (4x4 grid).
```
smuppand: Update the documentation after correcting the helper signature and supported session layouts.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/README.md
smuppand: Make the introductory test list consistently describe the 480p default.

smuppand:
@nitinn22  Please fix the shellcheck failed workflow.

nitinn22:
addressed comments and shell check failures.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1644,6 +1813,18 @@ camera_setup_wayland_environment() {
     if command -v weston_pick_env_or_start >/dev/null 2>&1; then
       log_info "No Wayland socket found, attempting to start Weston..."
       if weston_pick_env_or_start "$test_name"; then
+        # Track that we started Weston
+        GSTREAMER_STARTED_WESTON=1
+        export GSTREAMER_STARTED_WESTON
+        
+        # Register cleanup trap if not already registered
+        if [ -z "${GSTREAMER_WESTON_CLEANUP_REGISTERED:-}" ]; then
+          trap 'camera_cleanup_wayland_environment' EXIT INT TERM
```
smuppand: Remove trap registration from the library. Expose the cleanup helper and invoke it from the runner’s existing composed cleanup function.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1682,3 +1863,534 @@ camera_setup_wayland_environment() {
...
+# camera_cleanup_wayland_environment
+# Cleanup function to restore Weston state if it was started by camera_setup_wayland_environment
+# Called automatically via trap on EXIT, INT, TERM
+camera_cleanup_wayland_environment() {
+  # Only cleanup if we started Weston
+  if [ "${GSTREAMER_STARTED_WESTON:-0}" -eq 1 ]; then
+    log_info "Cleaning up Weston (started by camera_setup_wayland_environment)"
+    
+    # Use lib_display.sh restoration pattern if available
+    if command -v weston_restore_runtime >/dev/null 2>&1; then
```
smuppand: Use the captured initial state. Call weston_stop when this helper started Weston,  use weston_restore_runtime only when restoring a compositor previously stopped by the test.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -597,111 +668,961 @@ run_decode_test() {
...
+  
+  # Validate log
+  if ! gstreamer_validate_log "$test_log" "$testname"; then
+    log_fail "$testname: FAIL (GStreamer errors detected)"
+    fail_count=$((fail_count + 1))
+    return 1
+  fi
+  
+  # Verify rotation using shared helper (90° rotation should swap dimensions: 3840x2160 → 2160x3840)
+  if ! gstreamer_verify_rotation "$testname" "$output_file" "3840" "2160" "90"; then
```
smuppand: gstreamer_verify_rotation expects 2160x3840, but line 1384 immediately validates the same output as 3840x2160. The helper also returns success when gst-discoverer is missing or dimensions cannot be parsed.

Require usable evidence and validate once against the rotated 2160x3840 dimensions.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1682,3 +1863,534 @@ camera_setup_wayland_environment() {
...
+# gstreamer_build_smart_encode_pipeline
+# Build pipeline for HEVC Smart Encode with dual camera streams
+# Args: output_file
+# Returns: pipeline string or empty on error
+gstreamer_build_smart_encode_pipeline() {
+  output_file="$1"
+  
+  [ -n "$output_file" ] || return 1
+  
+  printf 'qtiqmmfsrc camera=0 noise-reduction=2 video_0::extra-buffers=20 video_0::type=preview name=camsrc video_0::type=video ! video/x-raw,format=NV12,width=1280,height=720,framerate=30/1 ! queue ! scb.sink qtismartvencbin default-gop=30 max-gop=600 smart-framerate=true smart-gop=false max-bitrate=4200000 name=scb encoder=v4l2h265enc ! queue ! h265parse ! queue ! mp4mux ! queue ! filesink location=%s camsrc. ! video/x-raw,format=NV12,width=640,height=480,framerate=15/1 ! queue ! scb.sink_ctrl' \
```
smuppand: The FR74943 test and documentation claim smart FPS and smart bitrate, but the pipeline only enables smart-framerate=true. Add the supported smart-bitrate=true property and verify the resulting behavior.

nitinn22: Although we can add smart-bitrate=true, there is no direct way to validate the result from the file. we ensure that the pipeline runs successfully and bitstream is dumped, idea of the feature is to allocate bits and frames dynamically based on the scene in the realtime from camera.

nitinn22:
### `run.sh`

- Updated test names and function comments
- Changed resolution parameters from 3840x2160 to 1920x1080
- Enhanced pipeline controls with VBR/CBR modes and QP parameters
- Updated validation logic for new resolutions
- Updated help text to reflect new test names

### `README.md`

- Updated test descriptions and names
- Changed resolution specifications from 4K/UHD to 1080p
- Added detailed feature lists for both tests
- Updated pipeline documentation with new controls

Runner/utils/lib_gstreamer.sh
smuppand: > Although we can add smart-bitrate=true, there is no direct way to validate the result from the file. we ensure that the pipeline runs successfully and bitstream is dumped, idea of the feature is to allocate bits and frames dynamically based on the scene in the realtime from camera.

The file does not need to prove the internal bitrate-allocation algorithm by itself, but the testcase must at least enable the feature it claims to exercise. Without smart-bitrate=true, this remains only a smart-framerate test

nitinn22: added smart-bitrate=true in command

Runner/utils/lib_gstreamer.sh
```diff
@@ -1357,43 +1370,217 @@ run_pipeline_with_logs() {
...
     console_log="${logdir}/${name}_console.log"
     gst_debug_log="${logdir}/${name}_gst_debug.log"
+    dmesg_dir="${logdir}/dmesg"
+    dmesg_log="${dmesg_dir}/${name}_dmesg.log"
+
+    # Create dmesg directory if it doesn't exist
+    mkdir -p "$dmesg_dir" 2>/dev/null || true
+
+    # Capture dmesg line count before test execution
+    dmesg_before_lines=$(dmesg | wc -l 2>/dev/null || echo 0)
```
smuppand: Please remove the direct dmesg invocation. Repository tests must use scan_dmesg_errors with retained dmesg_snapshot.log and dmesg_errors.log artifacts.

The current line-count approach is race-prone and can lose evidence when dmesg access is restricted, cleared, or wrapped. DRC transition validation should rely on ordered pipeline caps rather than unrelated kernel messages.

smuppand (changes requested):
Follow-up on UVC thread

This remains unresolved. Missing frame/FPS evidence still only produces a warning at line 1437, after which timeout can be accepted as PASS. Require positive frame or buffer activity and fail when it is absent.

Follow-up on DRC thread

This remains unresolved. The implementation only finds 1080p and 720p anywhere in combined console and dmesg output. It still does not prove an ordered 1080p→720p caps transition or require frame activity after the second caps event.

Follow-up on concurrency thread

This is only partially resolved. The decoder count is now strict, but it counts element names from verbose output. That proves construction, not that every branch negotiated caps and produced buffers. Require per-branch negotiated and active-buffer evidence.

Follow-up on Weston cleanup thread

This remains unresolved. run.sh cleanup still does not call camera_cleanup_wayland_environment. Also, weston_was_running is reset during every setup call, which can overwrite the suite’s original compositor state.

Follow-up on artifact-validation thread

This remains unresolved. gstreamer_validate_encode_output still returns success when gst-discoverer-1.0 is unavailable. Advanced encode tests must not PASS without validating codec, container, resolution, and duration.

nitinn22:
- __UVC Validation__: UVC tests now require positive frame/buffer activity evidence and fail when absent, preventing timeout acceptance without proof of actual video processing.

- __DRC Validation__: DRC tests now verify ordered 1080p→720p caps transition in console log AND require frame activity after the 720p caps event to prove the resolution change completed successfully.

- __Concurrency Validation__: Concurrency tests now verify each decoder branch individually for both caps negotiation and buffer activity, ensuring all parallel decode streams are actually working, not just constructed.

- __Artifact Validation__: Advanced encode tests already require gst-discoverer-1.0 and fail immediately if unavailable, preventing PASS results without validating codec, container, resolution, and duration.

- __Weston Cleanup__: Camera_Tests cleanup function already calls camera_cleanup_wayland_environment, and Weston state is captured once in GSTREAMER_WESTON_INITIAL_STATE (persists across calls), ensuring proper compositor restoration.

Will test code and update the PR.

nitinn22:
@smuppand , after testing the caps and buffer validation changes. the scripts seem fine from implementation standpoint but the logs are failing in validation. seeing sample stdout below. But the expectation on screen is fine wrt all 8 videos are playing fine, the logs arent coming properly for these validation, can we instead make these warning and pass them based on strict check of 8 session initiation as they are L1 tests?
[INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: Verifying per-branch caps negotiation and buffer activity...
[INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p:   ✓ v4l2h265dec0: Caps negotiated and buffers active
[FAIL] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p:   ✗ v4l2h265dec1: No caps negotiation detected
[INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p:   ✓ v4l2h265dec2: Caps negotiated and buffers active
[INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p:   ✓ v4l2h265dec3: Caps negotiated and buffers active
[INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p:   ✓ v4l2h265dec4: Caps negotiated and buffers active
[FAIL] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p:   ✗ v4l2h265dec5: No caps negotiation detected
[INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p:   ✓ v4l2h265dec6: Caps negotiated and buffers active
[FAIL] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p:   ✗ v4l2h265dec7: No caps negotiation detected
[FAIL] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: FAIL - 3/8 branches failed validation

smuppand:
> Srikanth Muppandam (@smuppand) , after testing the caps and buffer validation changes. the scripts seem fine from implementation standpoint but the logs are failing in validation. seeing sample stdout below. But the expectation on screen is fine wrt all 8 videos are playing fine, the logs arent coming properly for these validation, can we instead make these warning and pass them based on strict check of 8 session initiation as they are L1 tests? [INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: Verifying per-branch caps negotiation and buffer activity... [INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: ✓ v4l2h265dec0: Caps negotiated and buffers active [FAIL] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: ✗ v4l2h265dec1: No caps negotiation detected [INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: ✓ v4l2h265dec2: Caps negotiated and buffers active [INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: ✓ v4l2h265dec3: Caps negotiated and buffers active [INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: ✓ v4l2h265dec4: Caps negotiated and buffers active [FAIL] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: ✗ v4l2h265dec5: No caps negotiation detected [INFO] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: ✓ v4l2h265dec6: Caps negotiated and buffers active [FAIL] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: ✗ v4l2h265dec7: No caps negotiation detected [FAIL] 2026-08-20 05:34:57 - H265_Decode_Concurrency_8x480p: FAIL - 3/8 branches failed validation

Please do not downgrade the per-branch caps/buffer failures to warnings. Even for L1 coverage, a PASS needs machine-verifiable evidence that every requested decode branch became active, screen observation and decoder-name counts can produce false passes.

Please push the tested implementation first. Then make the evidence collection reliable, for example by enabling deterministic GStreamer verbosity/debug for the explicitly named branches and validating caps plus buffer/frame activity for each branch. If that evidence cannot be collected on a supported target, the testcase should be skipped with a clear capability reason, not passed.

nitinn22:
> Please do not downgrade the per-branch caps/buffer failures to warnings. Even for L1 coverage, a PASS needs machine-verifiable evidence that every requested decode branch became active, screen observation and decoder-name counts can produce false passes.
> 
> Please push the tested implementation first. Then make the evidence collection reliable, for example by enabling deterministic GStreamer verbosity/debug for the explicitly named branches and validating caps plus buffer/frame activity for each branch. If that evidence cannot be collected on a supported target, the testcase should be skipped with a clear capability reason, not passed.

Sure, pushed the changes. for now although the test is running it would fail due to this strict check, working with dev counterparts meanwhile to see if with log levels or any other verbosity we can validate these exactly and make them pass.

nitinn22:
Shell check failures are fixed.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -637,130 +668,986 @@ run_decode_test() {
...
+    return 1
+  fi
+  
+  log_info "Pipeline: $pipeline"
   
-  # Check if WebM file already exists
-  if [ -f "$vp9_clip_webm" ]; then
-    log_info "VP9 WebM clip already exists: $vp9_clip_webm"
+  # Use run_pipeline_with_logs with extended validation
+  if run_pipeline_with_logs "$testname" "$pipeline" "$OUTDIR" "$duration" "uvc" "1"; then
```
smuppand, line 840: adds a hardware pipeline path through run_pipeline_with_logs, but its shared timeout runner has neither a hard-kill escalation after SIGINT nor a safe fallback when timeout is unavailable. A wedged V4L2 ioctl can therefore consume the whole LAVA action timeout. Please make the shared runner enforce a portable SIGINT grace period followed by SIGKILL, and reject an unbounded fallback.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -637,130 +668,986 @@ run_decode_test() {
...
+  
+  # Check if display is connected (required for waylandsink)
+  if ! check_display_connected; then
+    log_skip "$testname: No display connected (required for waylandsink)"
+    skip_count=$((skip_count + 1))
+    return 1
+  fi
+  
+  # Set up Wayland display using shared function
+  if ! camera_setup_wayland_environment "uvc-preview"; then
```
smuppand, line 819: The suite calls camera_setup_wayland_environment(), which starts Weston when no socket exists. This PR simultaneously adds gstreamer_adopt_wayland_session_only() specifically for this suite, but never uses it. Video encode/decode must adopt an existing compositor session and SKIP if absent; it must not start or stop a compositor as an implicit side effect.

Runner/utils/lib_gstreamer.sh
```diff
@@ -2076,3 +1907,617 @@ camera_setup_wayland_environment() {
...
+  log_info "$testname: ✓ File size acceptable ($file_size bytes)"
+  
+  # 4. REQUIRE gst-discoverer-1.0 to verify container, codec, resolution, duration
+  if ! command -v gst-discoverer-1.0 >/dev/null 2>&1; then
+    log_fail "$testname: FAIL (gst-discoverer-1.0 required for validation but not available)"
+    return 1
+  fi
+  
+  discover_log="${output_file}.discover.log"
+  if ! gst-discoverer-1.0 "$output_file" >"$discover_log" 2>&1; then
```
smuppand: runs gst-discoverer-1.0 without a timeout, and the rotation verifier repeats the same pattern. A malformed or problematic generated artifact can hang validation indefinitely after the pipeline finishes. Please run discovery through the bounded hard-kill media runner and report timeout as a test failure.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -637,130 +668,986 @@ run_decode_test() {
...
+  dest_path="$2"
+  
+  # Check if clip already exists
+  if [ -f "$dest_path" ]; then
+    log_info "Clip already exists: $clip_name"
+    return 0
+  fi
+  
+  # Try to get from provided local path first
+  if [ -n "$clipPath" ] && [ -f "$clipPath/$clip_name" ]; then
```
smuppand, line 1293: only treats VIDEO_CLIP_PATH as a directory, while the YAML documents dir/file/archive inputs. A supplied archive or clip file is ignored and the test falls back to download or skip. Please implement the documented file/archive cases or restrict the interface and documentation to directory-only.

smuppand (changes requested):

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -637,130 +668,986 @@ run_decode_test() {
...
+    fail_count=$((fail_count + 1))
+    return 1
+  fi
+}
 
-# Run encode tests (skip VP9 as it doesn't support encoding in this test)
-if [ "$testMode" = "all" ] || [ "$testMode" = "encode" ]; then
+# -------------------- H.264 Concurrency Decode Test --------------------
+# Test: H264_Decode_Concurrency_8x480p (PR43865)
+run_h264_concurrency_test() {
```
smuppand: starts one of three largely duplicated concurrency wrappers. H.264, H.265, and MJPEG repeat capability checks, clip staging, Wayland setup, logging, counters, and execution, so fixes will drift between cases. Please use one data-driven run_concurrency_decode_case helper with codec, clip, decoder, and session-count inputs, leaving only genuinely codec-specific pipeline construction in the shared builder.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -637,130 +668,986 @@ run_decode_test() {
...
+    return 0
+  else
+    fail_count=$((fail_count + 1))
+    return 1
+  fi
+}
+
+# -------------------- HEVC Cyclic IR Encode Test --------------------
+# Test: HEVC_Encode_1080p_Cyclic_IR (FR82773) - Downstream only
+run_cyclic_ir_encode_test() {
```
smuppand, line 1225: begins another camera-encode wrapper with the same lifecycle as the cyclic-IR, slice-MB, and rotation cases. The copied setup, launch, validation, cleanup, and counter handling will drift as cases grow. Please use one parameterized camera-encode runner with a case table for codec, dimensions, controls, expected output, and an optional post-validator such as rotation.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -637,130 +668,986 @@ run_decode_test() {
...
+# Route to appropriate test based on test type
+case "$testType" in
+  basic|all)
+    # Prepare and run basic tests
+    prepare_basic_clips
+    run_basic_tests
+    
+    # If test type is 'all', prepare and run advanced tests too
+    if [ "$testType" = "all" ]; then
+      prepare_advanced_clips
```
smuppand, line 1465: makes --test-type all run the basic matrix plus nine advanced cases serially. With the default 30-second duration, pipeline allowances alone are about 550 seconds before clip download and setup. Please define the cases as a configurable test plan with a documented LAVA budget and a limit/selection mechanism for CI.

smuppand (changes requested):

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
nitinn22: CI can run as 10 second option if required, it is configurable.

Runner/utils/lib_gstreamer.sh
```diff
@@ -929,22 +931,24 @@ gstreamer_run_gstlaunch_timeout() {
...
+  if ! command -v timeout >/dev/null 2>&1; then
+    log_fail "timeout command required but not available - cannot enforce bounded execution"
+    return 1
+  fi
+
+  if [ "$secs" -gt 0 ] 2>/dev/null; then
+    # Send SIGINT to trigger EOS via -e flag, with 10s grace period before SIGKILL
+    # This prevents wedged V4L2 ioctls from consuming entire LAVA action timeout
+    # shellcheck disable=SC2086
+    timeout --signal=INT --kill-after=10 "$secs" "$GSTBIN" $GSTLAUNCHFLAGS $pipe
```
smuppand, line 944: This replaces the existing portable gstreamer_run_bounded path with GNU-specific timeout --signal=INT --kill-after=10.  Also i recommend to use run_with_timeout.

Minimal Yocto images commonly provide BusyBox timeout, which does not support these options, so this shared helper will now fail not only video tests but existing audio, camera, display, and graphics callers. Keep gstreamer_run_bounded, or extend it to provide the  INT/grace/KILL policy while preserving capability detection and a classified no-timeout result.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1761,33 +1780,221 @@ run_pipeline_with_logs() {
...
-        return 0
+    if [ "$playing" -eq 0 ]; then
+        log_fail "${name} FAIL: Pipeline never reached PLAYING state"
+        return 1
     fi
+    
+    # Check 2: No ERROR messages (original check)
+    error_present=$(grep -c "ERROR:" "$console_log" || true)
+    
+    # Extended validation for timeout-based tests (if test_type provided)
```
smuppand, line 1793: The extended checks search console_log for caps, per-decoder instance names, and buffer activity, but run_pipeline_with_logs() redirects  GStreamer debug output to GST_DEBUG_FILE. At the configured GST_DEBUG=2, normal gst-launch console output does not reliably contain caps     negotiation or per-element buffer evidence. 

Healthy UVC, DRC, and concurrency runs will therefore fail. Capture a bounded, explicitly  configured caps/activity debug stream and validate that retained debug artifact instead.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -637,111 +756,756 @@ run_decode_test() {
...
+      return 0
+    else
+      log_warn "Failed to copy $clip_name from local path"
     fi
   fi
-fi
+  
+  # Try to download from URL
+  log_info "Attempting to download $clip_name from URL..."
+  if extract_tar_from_url "$clipUrl" "$OUTDIR"; then
```
smuppand, line 1305: extract_tar_from_url() accepts one argument and extracts into ${LOG_DIR:-.}; its second "$OUTDIR" argument is ignored. This suite sets LOG_DIR to …/logs, then checks only the current directory and $OUTDIR, so automatically downloaded clips are not found and advanced tests skip despite successful extraction. 

Either extend the helper with a documented output-directory argument or search the helper’s actual extraction location.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -131,9 +135,10 @@ cleanup() {
   if ! pkill -P "$$" -x gst-launch-1.0 >/dev/null 2>&1; then
     pkill -x gst-launch-1.0 >/dev/null 2>&1 || true
   fi
-  # The element-probe cache is backed by files under TMPDIR; do not leave them.
-  if command -v gstreamer_reset_element_cache >/dev/null 2>&1; then
-    gstreamer_reset_element_cache
+
```
smuppand, line 138: reports 371 whitespace errors across the PR, including many new trailing-whitespace lines and a new blank line at EOF. Please clean the entire patch before merge.

Runner/utils/lib_gstreamer.sh
```diff
@@ -2076,3 +2305,692 @@ camera_setup_wayland_environment() {
...
+  
+  # REQUIRE timeout command for bounded discovery
+  if ! command -v timeout >/dev/null 2>&1; then
+    log_fail "$testname: FAIL (timeout command required for bounded discovery)"
+    return 1
+  fi
+  
+  discover_log="${output_file}.discover.log"
+  # Run gst-discoverer with timeout and SIGKILL escalation (30s timeout, 10s grace)
+  if ! timeout --signal=INT --kill-after=10 30 gst-discoverer-1.0 "$output_file" >"$discover_log" 2>&1; then
```
smuppand, line 2721: if ! command; then discover_rc=$? stores the status of !, not gst-discoverer-1.0; it is always 0 inside this branch. Timeout and failure diagnostics are therefore wrong. Run the command first, save its status, then branch on that saved value. The identical issue exists at lib_gstreamer.sh:2926.

smuppand (changes requested):

nitinn22:
1. __Whitespace__: Removed trailing whitespace (run.sh line 138)

2. __gst-discoverer timeout__: Fixed `if ! cmd; then rc=$?` capturing wrong exit status (lib_gstreamer.sh line 2722). Now uses portable `gstreamer_run_bounded()` and captures status before conditional check.

3. __extract_tar_from_url__: Fixed clip search path - now checks `${LOG_DIR:-}` where files are actually extracted, not just current dir and $OUTDIR (run.sh lines 1305-1321).

All changes maintain POSIX compliance and BusyBox compatibility.

Runner/utils/lib_gstreamer.sh
```diff
@@ -929,22 +934,20 @@ gstreamer_run_gstlaunch_timeout() {
...
   gstreamer_print_cmd_multiline "$pipe"
 
+  # Use gstreamer_run_bounded for portable timeout handling
+  # Send SIGINT to trigger EOS via -e flag, with GST_KILL_GRACE period before SIGKILL
+  # This prevents wedged V4L2 ioctls from consuming entire LAVA action timeout
   # shellcheck disable=SC2086
-  # Send SIGINT instead of SIGTERM to trigger EOS via -e flag
   gstreamer_run_bounded "$secs" INT "$GSTBIN" $GSTLAUNCHFLAGS $pipe
-  gstlaunch_rc=$?
+  gst_rc=$?
```
smuppand: This refactor removes the only read of GST_BOUNDED_RAW_RC. ShellCheck consequently reports SC2034 on the shared helper assignment, and the required PR ShellCheck job is failing. Preserve the raw-status diagnostic, especially the SIGKILL-escalation distinction, or remove/rework the variable consistently across
the helper and callers.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1761,33 +1779,225 @@ run_pipeline_with_logs() {
...
+    # Check 2: No ERROR messages (original check)
+    error_present=$(grep -c "ERROR:" "$console_log" || true)
+    
+    # Extended validation for timeout-based tests (if test_type provided)
+    if [ -n "$test_type" ]; then
+        log_info "${name}: ✓ PLAYING state reached"
+        
+        # Check 2b: Negotiated video caps observed (extended check)
+        # Search gst_debug_log since GST_DEBUG_FILE redirects debug output there
+        if ! grep -q "caps.*video/x-raw" "$gst_debug_log"; then
```
smuppand: The extended path requires caps in gst_debug_log, but the default GST_DEBUG=2 generally only records warnings/errors there. Negotiated caps produced by
gst-launch -v are written to console_log. Healthy UVC, DRC, and concurrency pipelines can therefore fail before their mode-specific validation. Validate both
retained logs or enable a targeted caps debug category explicitly.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1761,33 +1779,225 @@ run_pipeline_with_logs() {
...
+                    log_fail "${name}: FAIL - Invalid transition order (1080p at line $first_1080p_line, 720p at line $first_720p_line)"
+                    log_fail "${name}:   Expected: 1080p → 720p (1080p line < 720p line)"
+                    return 1
+                fi
+                
+                log_info "${name}: ✓ Ordered transition verified (1080p → 720p)"
+                
+                # REQUIRE frame activity after the 720p caps event (second resolution)
+                # Extract all lines after the 720p caps event
+                lines_after_720p=$(tail -n +$((first_720p_line + 1)) "$console_log" || true)
```
smuppand: first_720p_line is a line number from gst_debug_log, but it is applied to console_log here. These logs have unrelated line numbering, so the claimed post-
transition frame check can falsely pass or fail. Locate the transition and subsequent activity in the same retained evidence stream.

Runner/utils/lib_gstreamer.sh
```diff
@@ -1761,33 +1779,225 @@ run_pipeline_with_logs() {
...
+                    # Look for caps events mentioning this specific decoder instance
+                    if ! grep -q "${decoder_instance}.*caps.*video/x-raw\|caps.*video/x-raw.*${decoder_instance}" "$console_log"; then
+                        log_fail "${name}:   ✗ ${decoder_instance}: No caps negotiation detected"
+                        failed_branches=$((failed_branches + 1))
+                        continue
+                    fi
+                    
+                    # Check if this decoder instance has buffer activity evidence
+                    # Look for buffer/frame activity mentioning this specific decoder instance
+                    if ! grep -q "${decoder_instance}.*buffer\|${decoder_instance}.*frame\|chain.*${decoder_instance}" "$console_log"; then
```
smuppand: gst-launch -v does not emit per-buffer activity for each decoder into console_log. At the default debug level this grep cannot prove active buffers and can fail every healthy concurrency run. Instrument each branch with measurable activity or capture a targeted debug stream containing branch-specific buffer  evidence.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
smuppand: Still not resolved.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -131,9 +135,10 @@ cleanup() {
...
+  # Clean up Wayland/Weston if we started it
+  if command -v camera_cleanup_wayland_environment >/dev/null 2>&1; then
```
smuppand: Before extending this EXIT cleanup, remove the name-wide pkill fallback immediately above. When no direct child is found, pkill -x gst-launch-1.0
terminates unrelated pipelines on the target. Track the PIDs or process groups started by this suite and stop only those.

Runner/utils/lib_gstreamer.sh
```diff
@@ -2076,3 +2308,697 @@ camera_setup_wayland_environment() {
...
+    log_fail "$testname: FAIL (timeout command required for bounded discovery)"
+    return 1
+  fi
+  
+  log_info "$testname: Verifying ${rotation_degrees}° rotation..."
+  
+  # Get discoverer output and save to log file for debugging
+  discover_log="${output_file}.rotation_discover.log"
+  # Run gst-discoverer with timeout and SIGKILL escalation (30s timeout, 10s grace)
+  if ! timeout --signal=INT --kill-after=10 30 gst-discoverer-1.0 "$output_file" >"$discover_log" 2>&1; then
```
smuppand: Inside `if ! timeout ...; then`, `$?` is the status of `!`, so discover_rc is always zero. Timeout and command-failure diagnostics are incorrect. Run the bounded command, save its status, and branch on that value. Reuse gstreamer_run_bounded rather than invoking timeout directly.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
```diff
@@ -637,111 +756,761 @@ run_decode_test() {
...
+    # If test type is 'all', prepare and run advanced tests too
+    if [ "$testType" = "all" ]; then
+      prepare_advanced_clips
+      run_advanced_tests
     fi
-  done
-fi
+    ;;
+    
+  uvc)
```
smuppand: The global dependency check requires curl and tar before dispatch, although UVC and advanced-encode use no downloaded clips. Minimal Yocto images can skip
valid camera coverage unnecessarily. Gate clip-fetch dependencies only for modes that require downloaded assets.

Runner/suites/Multimedia/GSTreamer/Video/Video_Encode_Decode/run.sh
smuppand: Still open.

VIDEO_CLIP_PATH remains directory-only, while the YAML advertises directory, file, and archive inputs. Either support the documented file/archive forms through the existing local-clip preparation helper or restrict the YAML contract to directories.

smuppand (changes requested):

PR #522, opened 2026-07-27
add: audio sanity check to kernel plan

vnarapar:
Add new testcase to check audio soundcard registration to kernel qcom-next testplan

- Minor update to openCV test parameter to remove duplicate run.sh


smuppand (approved):
LGTM

PR #525, opened 2026-07-28
fix: prevent template injection in zstd archive workflow

vnarapar:
The Create .tar.zst archive step in zstd.yml used
${{ steps.version.outputs.tag_name }} directly inside a run: shell block. This allows template injection since the value is derived from GITHUB_REF which is attacker-controllable via a crafted git tag.

Pass the value through an env: variable instead and reference it as $TAG_NAME in the shell script to prevent direct template expansion into the run context.

This resolves the template-injection findings reported by the zizmor GitHub Actions security scan.


smuppand (approved):
LGTM

PR #526, opened 2026-07-28
Add distro package support for Sensors validation

smuppand:
Extend the Sensors testcase to support dependency discovery and package recovery on distro-based Linux images.

The testcase previously depended on the required Qualcomm sensor utilities already being present in the target image. This change adds package-provider mappings and updates the runner so the required sensor packages can be resolved consistently on supported distributions.


Vkenjam (approved):
LGTM

bhargav0610 (approved):

PR #528, opened 2026-08-03
Run URM Tests through Debian Packages

kartnema:
On platforms which use Debian builds, like Glymur, enable test case runs by installing the appropriate testing package for URM: "userspace-resource-manager-tests". This package installs the necessary binaries and configs needed for testing.

The run.sh script already has instructions for running the tests, fetching and reporting the results in an appropriate manner.


Runner/suites/Performance/userspace-resource-manager/run.sh
```diff
@@ -27,9 +27,23 @@ if [ -z "${__INIT_ENV_LOADED:-}" ]; then
...
 
+# Optional generic package-set recovery.
+# This must be a clean no-op when no package-set mapping exists for the active OS/provider.
+if [ -f "$TOOLS/lib_pkg_provider.sh" ]; then
+    # shellcheck disable=SC1091
+    . "$TOOLS/lib_pkg_provider.sh"
+
+    if ! pkg_ensure_package_set urm; then
+        log_skip "$TESTNAME SKIP - required package set is not available: urm"
+        echo "$TESTNAME SKIP" >"$RES_FILE"
```
smuppand: "RES_FILE" and "TESTNAME" have not been initialized when this failure path executes. they are assigned later at lines 55–58. If package recovery fails, the redirect targets an empty filename, no ".res" file is generated. 

Move the complete package-recovery block below:
```
TESTNAME="userspace-resource-manager"
test_path="$(find_test_case_by_name "$TESTNAME")"
cd "$test_path" || exit 1
RES_FILE="./${TESTNAME}.res"
```

Runner/suites/Performance/userspace-resource-manager/run.sh
```diff
@@ -27,9 +27,23 @@ if [ -z "${__INIT_ENV_LOADED:-}" ]; then
...
 # shellcheck disable=SC1090,SC1091
 . "$TOOLS/functestlib.sh"
 
+# Optional generic package-set recovery.
+# This must be a clean no-op when no package-set mapping exists for the active OS/provider.
+if [ -f "$TOOLS/lib_pkg_provider.sh" ]; then
+    # shellcheck disable=SC1091
+    . "$TOOLS/lib_pkg_provider.sh"
+
+    if ! pkg_ensure_package_set urm; then
```
smuppand: Every Debian/Ubuntu run may perform avoidable package-manager, network, and package-upgrade work, increasing runtime, external dependencies, and flakiness.

Recommended fix:
```
if ! pkg_ensure_required_package_set_present urm; then
    log_skip "$TESTNAME SKIP - failed to ensure required package set: urm"
    echo "$TESTNAME SKIP" >"$RES_FILE"
    exit 0
fi
```
The shared helper explicitly documents that it avoids package-manager/network work when the required package set is already complete and verifies the package set again after recovery.

smuppand (changes requested):
Please update the commit message to a more clear to reflect with the current changes.

Runner/suites/Performance/userspace-resource-manager/run.sh
```diff
@@ -43,6 +44,19 @@ test_path="$(find_test_case_by_name "$TESTNAME")"
 cd "$test_path" || exit 1
 RES_FILE="./${TESTNAME}.res"
 
+# Optional generic package-set recovery.
+# This must be a clean no-op when no package-set mapping exists for the active OS/provider.
+if [ -f "$TOOLS/lib_pkg_provider.sh" ]; then
+    # shellcheck disable=SC1091
+    . "$TOOLS/lib_pkg_provider.sh"
+
+    if ! pkg_ensure_required_package_set_present urm; then
```
smuppand, line 53: "pkg_ensure_required_package_set_present" treats a missing package-set mapping as failure. On Yocto/RPM/OPKG, where this PR intentionally provides no "urm" mapping, the helper first fails verification, "pkg_ensure_package_set" performs its no-op, and the final verification fails again. 
The runner then writes SKIP and exits instead of continuing with the packages already supplied by the image.

Recommended fix:
```
if [ -f "$TOOLS/lib_pkg_provider.sh" ]; then
    # shellcheck disable=SC1091
    . "$TOOLS/lib_pkg_provider.sh"

    if pkg_lookup_package_set urm >/dev/null 2>&1; then
        if ! pkg_ensure_required_package_set_present urm; then
            log_skip "$TESTNAME SKIP - failed to ensure required package set: urm"
            echo "$TESTNAME SKIP" >"$RES_FILE"
            exit 0
        fi
    else
        log_info "No URM package-set mapping for this OS/provider; using image-provided assets"
    fi
fi
```

Runner/config/pkg_command_map.conf
```diff
@@ -188,6 +188,21 @@ opkg:Sensors:sns_test=qcom-sensors-api qcom-sensors-core qcom-sensors-registry q
 debian:package-set:fastrpc=fastrpc-support fastrpc-tests
 ubuntu:package-set:fastrpc=fastrpc-support fastrpc-tests
 
+# ---------------------------------------------------------------------------
+# Qualcomm userspace-resource-manager package set.
+#
+# Used by tests that explicitly call:
+# pkg_ensure_package_set urm
```
smuppand, line 195: The comment documents direct use of "pkg_ensure_package_set" and "pkg_ensure_optional_package_set", but the current runner uses "pkg_ensure_required_package_set_present". Update this block to document the actual required-set caller and clarify that absent mappings are intentionally ignored by the runner

Runner/suites/Performance/userspace-resource-manager/run.sh
```diff
@@ -43,6 +44,19 @@ test_path="$(find_test_case_by_name "$TESTNAME")"
 cd "$test_path" || exit 1
 RES_FILE="./${TESTNAME}.res"
 
+# Optional generic package-set recovery.
```
smuppand, line 47: Package recovery currently runs before the test lock and before CLI parsing. This means concurrent invocations can both enter package installation, and even "--help" performs package verification or installation. Consider sourcing the provider here but moving the package recovery operation after argument parsing and lock acquisition. At minimum, ensure "--help" remains side-effect free

smuppand (changes requested):

Runner/suites/Performance/userspace-resource-manager/run.sh
kartnema: Moved pkg_ensure_required_package_set_present urm; later in the file post cli argument processing block.

Runner/suites/Performance/userspace-resource-manager/run.sh
```diff
@@ -133,6 +141,16 @@ while [ $# -gt 0 ]; do
     shift
 done
 
+if pkg_lookup_package_set urm >/dev/null 2>&1; then
```
smuppand: I have two points to get confirmation. Then we are good to merge.

1. Package recovery runs before the RUN_MODE=list branch at line 345. Therefore ./run.sh --list can invoke apt, install packages and trigger package post-install actions. 

Skip package recovery when RUN_MODE=list. Only ensure the package set after confirming that an approved test will actually execute.

2.  Guard with command -v pkg_lookup_package_set and command -v pkg_ensure_required_package_set_present, or make the library explicitly required on mapped distributions.

Runner/suites/Performance/userspace-resource-manager/run.sh
```diff
@@ -337,6 +345,27 @@ if [ "$RUN_MODE" = "list" ]; then
     exit 0
 fi
 
+# ---------- Ensure packages (deferred: skip when RUN_MODE=list) ----------
```
smuppand: Move this package-recovery block to immediately after CLI parsing, while retaining the RUN_MODE != list guard. It must execute before the service gate, configuration validation and preflight_bins. Continue using pkg_lookup_package_set and pkg_ensure_required_package_set_present

Runner/suites/Performance/userspace-resource-manager/run.sh
```diff
@@ -133,6 +141,27 @@ while [ $# -gt 0 ]; do
     shift
 done
 
+# ---------- Ensure packages (deferred: skip when RUN_MODE=list) ----------
+# Only recover the package set once we know an approved test will actually
+# execute.  Guard with command -v so the calls are safe even when
+# lib_pkg_provider.sh was not sourced or the OS/provider has no mapping.
+if [ "$RUN_MODE" != "list" ]; then
```
smuppand: Maybe my earlier comments didn't convey in right way. 

Package recovery is gated only by `RUN_MODE != list` it does not confirm that the value supplied through --bin is approved. Consequently, ./run.sh --bin typo or a missing --bin value can invoke package verification, network access, installation and package post-install actions before the runner later skips the unapproved test. 

Move approved_tests/is_approved before this block and validate ONE_BIN, or construct the approved run list before recovery. Package recovery should execute only for --all or a validated, approved --bin

Runner/suites/Performance/userspace-resource-manager/run.sh
```diff
@@ -150,6 +158,46 @@ is_approved() {
...
+# ---------- Ensure packages ----------
+# Run only when at least one approved test will actually execute:
+#   --all              => always recover
+#   --bin NAME         => recover only when NAME is non-empty and approved
+# This prevents a typo or missing --bin value from triggering network access,
+# package installation and post-install actions before the runner skips the
+# unapproved test.
+# Guard with command -v so the calls are safe even when lib_pkg_provider.sh
+# was not sourced or the OS/provider has no mapping.
+_pkg_recovery_needed=0
```
smuppand: Please remove trailing underscore before var/functions names.

